PR #85, opened 2023-02-03
Mypy annotations

sanjanag:
- Fixed mypy errors introduced in recent PRs
- Refactored `params` argument from `__init__()` to `find_issues()`
- Moved demo.ipynb to examples/
- Updated custom_issue_manager

### DIFF
--- a/examples/custom_issue_manager.py
+++ b/examples/custom_issue_manager.py
@@ -60,7 +60,6 @@ class CustomIssueManager(IssueManager):
         imagelab_info: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
     ) -> None:
-
         super().find_issues(**kwargs)
         assert params is not None
         assert imagelab_info is not None
--- a/examples/custom_issue_manager.py
+++ b/examples/custom_issue_manager.py
@@ -1,3 +1,5 @@
+from typing import Any, Dict, List, Optional
+
 import numpy as np
 import pandas as pd
 from PIL import Image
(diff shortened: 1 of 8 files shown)


src/cleanvision/issue_managers/image_property.py
```diff
@@ -207,7 +207,7 @@ def get_image_mode(image: Image) -> str:
         imarr = np.asarray(image)
         if len(imarr.shape) == 2 or (
             len(imarr.shape) == 3
-            and (np.diff(imarr.reshape(-1, 3).T, axis=0) == 0).all()  # type: ignore
+            and (np.diff(imarr.reshape(-1, 3).T, axis=0) == 0).all()
```
ulya-tkch, line 210: Does this pass mypy typecheck? np.diff for my local versions caused error as an untyped function (thats why `# type: ignore` was there)

ulya-tkch (approved):
If this passes strict mypy then LGTM!

src/cleanvision/issue_managers/image_property.py
sanjanag: It did pass for me on local and looks like it's passing on CI as well

examples/custom_issue_manager.py
```diff
@@ -16,39 +18,55 @@ class CustomIssueManager(IssueManager):
...
         self.params = {**self.params, **non_none_params}
 
     @staticmethod
-    def calculate_mean_pixel_value(image):
+    def calculate_mean_pixel_value(image: Image.Image) -> float:
         gray_image = image.convert("L")
         return np.mean(np.array(gray_image))
 
-    def get_scores(self, raw_scores):
+    def get_scores(self, raw_scores: "np.ndarray[Any, Any]") -> "np.ndarray[Any, Any]":
```
elisno, line 41: Based on the only place that this is called (within the class): Raw scores should be a list of floats, while the return value is an array (vector?) of floats.

```suggestion
    def get_scores(self, raw_scores: List[float]) -> npt.NDArray[np.float_]:
```

examples/custom_issue_manager.py
```diff
@@ -16,39 +18,55 @@ class CustomIssueManager(IssueManager):
...
         scores = np.array(raw_scores)
         return scores / 255.0
 
-    def mark_issue(self, scores, threshold):
+    def mark_issue(
+        self, scores: "np.ndarray[Any, Any]", threshold: float
+    ) -> "np.ndarray[Any, Any]":
```
elisno, line 47: The output of `get_scores` is provided as `scores` in this method. So we can just use the same type hint.
The element-wise comparisons return an array (vector?) of booleans.

```suggestion
        self, scores: npt.NDArray[np.float_], threshold: float
    ) -> npt.NDArray[np.bool_]:
```

examples/custom_issue_manager.py
```diff
@@ -1,3 +1,5 @@
+from typing import Any, Dict, List, Optional
+
 import numpy as np
 import pandas as pd
```
elisno, line 4: Need this import for my other suggestions:

```suggestion
import numpy as np
import numpy.typing as npt
import pandas as pd
```

elisno (approved):
Looks great! :shipit: 

There seems to be one type issue with the use of `np.mean` in
examples/custom_issue_manager.py.

I haven't looked into examples/run.py, but there are some mypy warnings there, which I'm sure can be addressed in a future PR.

PR #86, opened 2023-02-04
find_issues multiprocessing

clu0:
notebook for benchmarking in `notebook/multiproc_benchmark.ipynb`, did not include results due to file size, but got ~40x speedup with 48 cores


src/cleanvision/issue_managers/duplicate_issue_manager.py
```diff
@@ -89,14 +125,45 @@ def find_issues(
...
+        if n_jobs is None:
+            if psutil_exists:
+                n_jobs = psutil.cpu_count(logical=False)  # physical cores
+            if not n_jobs:
+                # either psutil does not exist
+                # or psutil can return None when physical cores cannot be determined
+                # switch to logical cores
+                n_jobs = multiprocessing.cpu_count()
+
+        if n_jobs == 1:
```
sanjanag, line 137: There's some repetition of code in this way of implementing. How I would do it is, create a base method that populates `result` dict, then create a wrapper around this base method for using in `multiprocessing.Pool()`
Base method would look like
```suggestion
        def compute_hash(path, to_compute, params):
            # logic
            return result{"path": path, "issue_type0": hash0, "issue_type1": hash1  }
```
Wrapper would look like
```suggestion
def compute_hash_wrapper(args):
    path = args["path"]
    to_compute = args["to_compute"]
    params = args["params"]
    return compute_hash(path, to_compute, params)
```

src/cleanvision/issue_managers/duplicate_issue_manager.py
```diff
@@ -89,14 +125,45 @@ def find_issues(
...
+            for path in tqdm(filepaths):
+                image = Image.open(path)
```
sanjanag, line 139: call the base method in this loop

src/cleanvision/issue_managers/duplicate_issue_manager.py
```diff
@@ -89,14 +125,45 @@ def find_issues(
...
+                        issue_type_hash_mapping[issue_type][hash] = [path]
+        else:
+            args = [
+                {"path": path, "to_compute": to_compute, "params": self.params}
+                for path in filepaths
+            ]
+            with multiprocessing.Pool(n_jobs) as p:
+                results = list(
+                    tqdm(
+                        p.imap_unordered(compute_hash, args, chunksize=10),
```
sanjanag, line 154: use the wrapper method here

src/cleanvision/issue_managers/duplicate_issue_manager.py
```diff
@@ -89,14 +125,45 @@ def find_issues(
...
+                        total=len(filepaths),
+                    )
+                )
+
+            results = sorted(results, key=lambda r: r["path"])
```
sanjanag, line 159: and then this would be the common code for collecting results from both n_jobs=1 and otherwise

src/cleanvision/issue_managers/duplicate_issue_manager.py
```diff
@@ -10,6 +10,41 @@
...
+# This package is not necessary, because we can always fall back to logical cores as the default
+try:
+    import psutil
+
+    psutil_exists = True
+except ImportError as e:  # pragma: no cover
+    psutil_exists = False
+
+
+def get_hash(image: Image, params: Dict[str, Any]) -> str:
```
sanjanag, line 25: Since this is required for multiprocessing, remove the duplicate method in class and use this method everywhere

src/cleanvision/issue_managers/image_property_issue_manager.py
```diff
@@ -15,6 +15,43 @@
...
+)
+
+# psutil is a package used to count physical cores for multiprocessing
+# This package is not necessary, because we can always fall back to logical cores as the default
+try:
+    import psutil
+
+    psutil_exists = True
+except ImportError as e:  # pragma: no cover
+    psutil_exists = False
```
sanjanag, line 34: Move this to `utils/utils.py`

src/cleanvision/utils/utils.py
```diff
@@ -37,8 +37,10 @@ def get_filepaths(
     a list of image filenames sorted numerically and alphabetically
     """
 
-    if not os.path.isdir(dir_path):
-        raise NotADirectoryError
+    # this check means we cannot have dir_path be something like this: /256_ObjectCategories/*/
```
sanjanag, line 40: can you write an appropriate check? The idea is to check if a valid path exists here

src/cleanvision/issue_managers/image_property_issue_manager.py
```diff
@@ -96,13 +134,40 @@ def find_issues(
...
+            if n_jobs is None:
+                if psutil_exists:
+                    n_jobs = psutil.cpu_count(logical=False)  # physical cores
+                if not n_jobs:
+                    # either psutil does not exist
+                    # or psutil can return None when physical cores cannot be determined
+                    # switch to logical cores
+                    n_jobs = multiprocessing.cpu_count()
+
+            if n_jobs == 1:
```
sanjanag, line 146: For this issue manager too, follow the same pattern of code as it would reduce duplicate code.
1. create a base method that would take in path, to_compute and any other required arguments
2. write a wrapper function around it to be used in `imap_unordered()`
3. Collect results

src/cleanvision/issue_managers/image_property_issue_manager.py
```diff
@@ -15,6 +15,43 @@
...
+try:
+    import psutil
+
+    psutil_exists = True
+except ImportError as e:  # pragma: no cover
+    psutil_exists = False
+
+
+def compute_scores(arg: Dict[str, Any]) -> Dict[str, Any]:
+    compute_functions = {
```
sanjanag, line 38: Pass `compute_functions` dict as an argument

sanjanag:
The speedup is great, I added a few comments on how to organize the code better to reduce duplication.  Other than that make sure you pass the black format and flake8 checks in CI.
To see those issues, run:
`python -m black src`
`flake8 --ignore=E501,E722,E401,W503 src tests --count --show-source --statistics`

src/cleanvision/utils/utils.py
clu0: not sure how to check this, because glob expressions are allowed to have no matches in the glob.glob function

I instead added a print warning if no images were found

src/cleanvision/issue_managers/image_property_issue_manager.py
clu0: does this make a big difference? I'm in favor of keeping this as is, instead of passing it successively through the compute function and the wrapper, because it is not used anywhere else

clu0:
there's a type error that seems to be caused by using `sorted`, having some trouble resolving it

src/cleanvision/issue_managers/image_property_issue_manager.py
sanjanag: I agree passing the dict is an overhead, can you make it a global variable in the module and access it? Since it is a static object, it doesn't make sense to have it in the method.

src/cleanvision/utils/utils.py
```diff
@@ -48,6 +67,9 @@ def get_filepaths(
         if len(filetype_images) == 0:
             continue
         filepaths += filetype_images
+
+    if len(filepaths) < 1:
```
sanjanag, line 71: This check is already there in `imagelab.py`. You can remove it.

src/cleanvision/utils/utils.py
```diff
@@ -3,6 +3,17 @@
...
 from typing import Dict, List, Any
 
+import multiprocessing
+
+# psutil is a package used to count physical cores for multiprocessing
+# This package is not necessary, because we can always fall back to logical cores as the default
+try:
+    import psutil
+
+    psutil_exists = True
```
sanjanag, line 13: capitalize `psutil_exists`

src/cleanvision/issue_managers/image_property_issue_manager.py
clu0: There's actually something subtle here: your approach of making it a global variable would work *only on Linux*, because only there are sub-processes started with the fork method.

On MacOS and Windows, processes are started with spawn, so they would not have access to global variables. In that case, we would need to pass in these functions as inputs to the subprocesses, so the current code would be necessary anyway. Thus I still think the best is just to leave as is.

sanjanag: I see, thanks for explaining.

sanjanag (approved):

PR #91, opened 2023-02-08
Added docstrings for Imagelab module, class, public methods and attributes

sanjanag:
Added docstrings for Imagelab module, class, public methods and attributes

----
:books: Documentation preview :books:: https://cleanvision--91.org.readthedocs.build/en/91/

### DIFF
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -475,10 +475,6 @@ class Imagelab:
             If `issue_types` is given, it implies number of top examples per issue type to visualize.
             If `image_files` is given, this argument will be ignored.
 
-        examples_per_issue : int, optional
-            Number of top examples per issue type to visualize
-            Used only when `issue_types` is given and `image_files` is empty
-
         cell_size : Tuple[int, int], optional
             Dimensions controlling the size of each image in the depicted image grid.
 
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -553,7 +553,7 @@ class Imagelab:
     def get_stats(self) -> Any:
         return self.info["statistics"]
 
-    def save(self, path: str) -> None:
+    def save(self, path: str, allow_overwrite: bool = False) -> None:
         """Saves this ImageLab instance into a folder at the given path.
         Your saved Imagelab should be loaded from the same version of the CleanVision package.
         This method does not save your image files.
@@ -562,14 +562,26 @@ class Imagelab:
         ----------
         path : str
             Path to folder where this Imagelab instance will be saved on disk.
-            **Caution:** If a folder already exists at this location, it will be overwritten!
+        allow_overwrite: bool, default=False
+            If set to True, existing files at `path` will be overwritten.
+
+        Raises
+        ------
+        ValueError
+            If `allow_overwrite` is set to False, and an existing path is specified for saving Imagelab instance.
         """
-        if os.path.exists(path):
-            print(
-                f"WARNING: Existing files will be overwritten by newly saved files at: {path}"
-            )
-        else:
+        path_exists = os.path.exists(path)
+        if not path_exists:
             os.mkdir(path)
+        else:
+            if allow_overwrite:
+                print(
+                    f"WARNING: Existing files will be overwritten by newly saved files at: {path}"
+                )
+            else:
+                raise ValueError(
+                    "Path already exists, specify a new path or set allow_overwrite=True"
+                )
 
         self._path = path
         object_file = os.path.join(self._path, OBJECT_FILENAME)
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -340,7 +340,7 @@ class Imagelab:
             You are presumably already aware of these in your dataset.
 
         examples_per_issue : int, default=4
-            Number of examples to show for issue type reported.
+            Maximum number of images to show for issue type reported. These are examples of the top-most severe instances of the issue in your dataset.
 
         verbosity : int, {1, 2, 3, 4}
             Increasing verbosity increases the detail of the report. Set this to 0 to report less information, or to 4 to report the most information.
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -105,15 +105,19 @@ class Imagelab:
         self._config: Dict[str, Any] = self._set_default_config()
         self._path = ""
 
-    def _get_filepaths(self, data_path, filepaths):
+    def _get_filepaths(
+        self, data_path: Optional[str], filepaths: Optional[List[str]]
+    ) -> List[str]:
+        filepaths = []
         if not data_path and not filepaths:
             raise ValueError(
                 "Please specify data_path or filepaths to check for issues."
             )
         elif data_path:
-            return get_filepaths(data_path)
+            filepaths = get_filepaths(data_path)
         elif filepaths:
-            return filepaths
+            filepaths = filepaths
+        return filepaths
 
     def _set_default_config(self) -> Dict[str, Any]:
         """Sets default values for various config variables used in Imagelab class
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -471,7 +471,7 @@ class Imagelab:
             Number of top examples per issue type to visualize
             Used only when issue_types is given and image_files is empty
 
-        cell_size : Tuple[int, int]
+        cell_size : Tuple[int, int], optional
             cell size of each image in the image grid when visualizing
 
         Examples
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -47,6 +47,7 @@ class Imagelab:
     Basic usage of Imagelab class
 
     .. code-block:: python
+
         from cleanvision.imagelab import Imagelab
         imagelab = Imagelab("FOLDER_WITH_IMAGES/")
         imagelab.find_issues()
@@ -58,7 +59,7 @@ class Imagelab:
     issues : pd.DataFrame
         Dataframe where each row corresponds to an image and columns specify which issues were detected in this image.
         It has two types of columns for each issue type:
-        1. <issue_type>_score - This column contains a quality-score for each image for a particular type of issue. 
+        1. <issue_type>_score - This column contains a quality-score for each image for a particular type of issue.
         Scores are between 0 and 1, lower values indicate images exhibiting more severe instances of this issue.
         2. <issue_type>_bool - This column indicates whether or not the issue_type is detected in each image (a binary decision rather than numeric score).
 
@@ -169,11 +170,13 @@ class Imagelab:
         To check for all default issue types use
 
         .. code-block:: python
+
             imagelab.find_issues()
 
         To check for specific issue types with default settings
 
         .. code-block:: python
+
             issue_types = {
                 "dark": {},
                 "blurry": {}
@@ -183,6 +186,7 @@ class Imagelab:
         To check for issue types with different hyperparameters.
 
         .. code-block:: python
+
             issue_types = {
                 "dark": {"threshold": 0.1},
                 "blurry": {}
@@ -328,7 +332,7 @@ class Imagelab:
             This must be subset of the issue types specified in ``imagelab.find_issues()``.
 
         num_top_issues : int, default=3
-            By default, this report will only cover the top most prevalent types of issues in the dataset, specified by this value. You can specify `issue_types` instead to override this and include specific types of issues in the report. 
+            By default, this report will only cover the top most prevalent types of issues in the dataset, specified by this value. You can specify `issue_types` instead to override this and include specific types of issues in the report.
 
         max_prevalence : float, default=0.5
             Between 0 and 1.
@@ -346,11 +350,13 @@ class Imagelab:
         Default usage
 
         .. code-block:: python
+
             imagelab.report()
 
         Report specific issue types
 
         .. code-block:: python
+
             issue_types = ["dark", "near_duplicates"]
             imagelab.report(issue_types=issue_types)
 
@@ -482,14 +488,17 @@ class Imagelab:
         To visualize random images from the dataset
 
         .. code-block:: python
+
             imagelab.visualize()
 
         .. code-block:: python
+
             imagelab.visualize(num_images=8)
 
         To visualize specfic images from the dataset
 
         .. code-block:: python
+
             image_files = ["./dataset/cat.png", "./dataset/dog.png", "./dataset/mouse.png"]
             imagelab.visualize(image_files=image_files)
 
@@ -497,6 +506,7 @@ class Imagelab:
         To visualize top examples of specific issue types from the dataset
 
         .. code-block:: python
+
             issue_types = ["dark", "odd_aspect_ratio"]
             imagelab.visualize(issue_types=issue_types)
 
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -326,7 +326,7 @@ class Imagelab:
             List of issue types to report
 
         num_top_issues : int, default=3
-            Number of top issues to report. It is ignored if issue_types is given
+            By default, this report will only cover the top most prevalent types of issues in the dataset, specified by this value. You can specify `issue_types` instead to override this and include specific types of issues in the report. 
 
         max_prevalence : float, default=0.5
             Between 0 and 1.
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -78,9 +78,7 @@ class Imagelab:
         }
 
     def list_default_issue_types(self) -> None:
-        """
-        Prints a list of all issue types checked by default if no issue types are specified in imagelab.find_issues()
-        """
+        """Prints a list of all issue types checked by default if no issue types are specified in imagelab.find_issues()"""
 
         print("Default issue type checked by Imagelab:\n")
         print(
@@ -89,8 +87,7 @@ class Imagelab:
         )
 
     def list_possible_issue_types(self) -> None:
-        """
-        Prints a list of all possible issue types that can be checked in the dataset.
+        """Prints a list of all possible issue types that can be checked in the dataset.
         It will also include custom added issue types.
         """
         print("All possible issues checked by Imagelab:\n")
@@ -102,16 +99,6 @@ class Imagelab:
     def _get_issues_to_compute(
         self, issue_types_with_params: Optional[Dict[str, Any]]
     ) -> Dict[str, Any]:
-        """
-
-        Parameters
-        ----------
-        issue_types_with_params
-
-        Returns
-        -------
-
-        """
         if not issue_types_with_params:
             to_compute_issues_with_params: Dict[str, Any] = {
                 issue_type.value: {}
@@ -290,6 +277,16 @@ class Imagelab:
         examples_per_issue: Optional[int] = None,
         verbosity: int = 1,
     ) -> None:
+        """Prints a summary report of issues found in the dataset with their example images from the dataset.
+
+        Parameters
+        ----------
+        issue_types :
+        num_top_issues
+        max_prevalence
+        examples_per_issue
+        verbosity
+        """
         assert isinstance(verbosity, int) and 0 < verbosity < 5
 
         user_supplied_args = locals()
@@ -305,14 +302,14 @@ class Imagelab:
         issue_summary = self.issue_summary[
             self.issue_summary["issue_type"].isin(computed_issue_types)
         ]
-        self.print_issue_summary(issue_summary)
+        self._pprint_issue_summary(issue_summary)
 
         self.visualize(
             issue_types=computed_issue_types,
             examples_per_issue=report_args["examples_per_issue"],
         )
 
-    def print_issue_summary(self, issue_summary: pd.DataFrame) -> None:
+    def _pprint_issue_summary(self, issue_summary: pd.DataFrame) -> None:
         issue_summary_copy = issue_summary.copy()
         issue_summary_copy.dropna(axis=1, how="all", inplace=True)
         issue_summary_copy.fillna("N/A", inplace=True)
@@ -403,7 +400,7 @@ class Imagelab:
 
         num_images : int, optional
             Number of images to randomly visualize from the dataset
-            Used only when image_files and issue_types are empty
+            Used only when image_files and issue_types are empty, otherwise this argument will be ignored
 
         examples_per_issue : int, optional
             Number of top examples per issue type to visualize
@@ -411,6 +408,31 @@ class Imagelab:
 
         cell_size : Tuple[int, int]
             cell size of each image in the image grid when visualizing
+
+        Examples
+        --------
+
+        To visualize random images from the dataset
+
+        .. code-block:: python
+        imagelab.visualize()
+
+        .. code-block:: python
+        imagelab.visualize(num_images=8)
+
+        To visualize specfic images from the dataset
+
+        .. code-block:: python
+        image_files = ["./dataset/cat.png", "./dataset/dog.png", "./dataset/mouse.png"]
+        imagelab.visualize(image_files=image_files)
+
+
+        To visualize top examples of specific issue types from the dataset
+
+        .. code-block:: python
+        issue_types = ["dark", "odd_aspect_ratio"]
+        imagelab.visualize(issue_types=issue_types)
+
         """
         if issue_types:
             for issue_type in issue_types:
@@ -456,7 +478,7 @@ class Imagelab:
     def load(
         cls: Type[TImagelab], path: str, data_path: Optional[str] = None
     ) -> TImagelab:
-        """Loads Imagelab from file.
+        """Loads Imagelab from given path.
         `path` is the path to the saved Imagelab, not pickle file.
         `data_path` is the path to image dataset previously used in Imagelab.
         If the `data_path` is changed, Imagelab will not be loaded as some of its functionalities depend on it.
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -374,10 +374,13 @@ class Imagelab:
         examples_per_issue: int = 4,
         cell_size: Tuple[int, int] = (2, 2),
     ) -> None:
-        """Visualization helper for images. This method has multipurpose and can be used for visualizaing:
+        """Visualization helper for images.
+
+        This method is multipurpose and can be used for visualizaing:
         1. Random images from the dataset
         2. Particular images with paths given in image_files
         3. Top examples of given issue_types found in the dataset
+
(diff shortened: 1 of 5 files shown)


src/cleanvision/imagelab.py
```diff
@@ -226,26 +365,26 @@ def report(
...
-        self.print_issue_summary(issue_summary)
+        self._pprint_issue_summary(issue_summary)
 
         self.visualize(
             issue_types=computed_issue_types,
             examples_per_issue=report_args["examples_per_issue"],
         )
 
-    def print_issue_summary(self, issue_summary: pd.DataFrame) -> None:
+    def _pprint_issue_summary(self, issue_summary: pd.DataFrame) -> None:
```
elisno, line 375: Typo when renaming this method?

```suggestion
    def _print_issue_summary(self, issue_summary: pd.DataFrame) -> None:
```

src/cleanvision/imagelab.py
```diff
@@ -320,9 +515,14 @@ def get_stats(self) -> Any:
         return self.info["statistics"]
 
     def save(self, path: str) -> None:
-        """Saves this ImageLab to file (all files are in folder at path/).
+        """Saves this ImageLab instance to the given path.
```
elisno, line 518:
```suggestion
        """Saves this ImageLab instance at the given path.
```

src/cleanvision/utils/utils.py
```diff
@@ -16,25 +16,25 @@
...
+     All image files with extension in TYPES are allowed.
+     Returns a sorted list of sorted filepaths
 
 
     Parameters
     ----------
-    dir_path: str (an attribute of ImageDataset Class)
-    a string represening the current working directory
+    dir_path: str
+    Path to the dir containing image files, can be relative or absolute path
```
elisno, line 31:
```suggestion
    dir_path: str
        Path to the dir containing image files, can be relative or absolute path
```

src/cleanvision/imagelab.py
sanjanag: My intention was to use pp similar to pretty print, but either one works

elisno: Ohh I see, then feel free to ignore my comment.

src/cleanvision/utils/utils.py
```diff
@@ -16,25 +16,25 @@
...
-    a string represening the current working directory
+    dir_path: str
+        Path to the dir containing image files, can be relative or absolute path
 
 
     Returns
     -------
-    sorted_names: list[str]
-    a list of image filenames sorted numerically and alphabetically
+    list of str
```
jwmueller, line 36:
```suggestion
    sorted_names: list[str]
```

jwmueller: Is the particular reason why you replaced the old part of this with "list of str" ?

The old version looks more polished and would be preferred style for rest of documentation IMO

src/cleanvision/imagelab.py
```diff
@@ -1,3 +1,6 @@
+"""Defines Imagelab class for finding all types of issues in image dataset
+"""
```
jwmueller, line 2:
```suggestion
"""
Imagelab is the core class in CleanVision for finding all types of issues in an image dataset.
The methods in this module should suffice for most use-cases,
but advanced users can get extra flexibility via the code in other CleanVision modules.
"""
```

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
 from cleanvision.utils.viz_manager import VizManager
 
+__all__ = ["Imagelab"]
+
 OBJECT_FILENAME = "imagelab.pkl"
 TImagelab = TypeVar("TImagelab", bound="Imagelab")
 
 
 class Imagelab:
+    """A single class to find all types of issues in image datasets.
```
jwmueller, line 34:
```suggestion
    """A single class to find all types of issues in image datasets. Imagelab detects issues in the raw image files themselves and thus can be useful in most computer vision tasks.
```

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
+
+    Parameters
+    ----------
+    data_path : str
+        Path to image files. Imagelab will recursively retrieve all image files from the specified path
+
```
jwmueller, line 40: Todo: add optional `file_paths` arg

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
+    Examples
+    --------
+
+    Basic usage of Imagelab class
+
+    .. code-block:: python
+        from cleanvision.imagelab import Imagelab
+        imagelab = Imagelab(data_path)
```
jwmueller, line 48:
```suggestion
        imagelab = Imagelab("FOLDER_WITH_IMAGES/")
```

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
+
+
+    Attributes
+    ----------
+    issues : pd.DataFrame
+        Dataframe containing scores on all issue types for all images in the dataset.
+        It has two types of columns for each issue type
+        1. <issue_type>_score - This column indicates the score of each image for that issue type. Lower scores imply
+        higher probability of issue_type being present in the image
+        2. <issue_type>_bool - This column indicates whether the issue_type is present in the image or not.
```
jwmueller, line 60:
```suggestion
    issues : pd.DataFrame
        Dataframe where each row corresponds to an image and columns specify which issues were detected in this image.
        It has two types of columns for each issue type:
        1. <issue_type>_score - This column contains a quality-score for each image for a particular type of issue. 
        Scores are between 0 and 1, lower values indicate images exhibiting more severe instances of this issue.
        2. <issue_type>_bool - This column indicates whether or not the issue_type is detected in each image (a binary decision rather than numeric score).
```

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
+    Basic usage of Imagelab class
+
+    .. code-block:: python
+        from cleanvision.imagelab import Imagelab
+        imagelab = Imagelab(data_path)
+        imagelab.find_issues()
+        imagelab.report()
+
+
+    Attributes
```
jwmueller, line 53: @sanjanag @elisno 
Currently Imagelab.issues contains column called: `<issue_type>_bool` for the hard decision on whether image contains issue or not. But Datalab.issues contains column called: `is_<issue_type>_issue` ?

We should unify these. I prefer the latter sort of naming convention where it is more clear what this column means just from the dataframe itself without user having to refer to the documentation.

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
+        Dataframe containing scores on all issue types for all images in the dataset.
+        It has two types of columns for each issue type
+        1. <issue_type>_score - This column indicates the score of each image for that issue type. Lower scores imply
+        higher probability of issue_type being present in the image
+        2. <issue_type>_bool - This column indicates whether the issue_type is present in the image or not.
+
+
+    issue_summary:
+        Dataframe containing summary of all issue types found.
+        Specifically, it shows num_images of each issue found in the dataset
```
jwmueller, line 68:
```suggestion
    issue_summary:
        Dataframe where each row corresponds to a type of issue and columns summarize the overall prevalence of this issue in the dataset. 
        Specifically, it shows the number of images detected with the issue.
```

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
+
+    info : dict
+        This is a nested dictionary that contains statistics on images or other useful information,
+        collected while checking for issues in the dataset.
```
jwmueller, line 72:
```suggestion
    info : dict
        Nested dictionary that contains statistics and other useful information about the dataset.
        Also contains additional information saved while checking for issues in the dataset.
```

src/cleanvision/imagelab.py
```diff
@@ -21,29 +24,82 @@
...
+
+    Raises
+    ------
+    ValueError
+        If not images are found in the data_path
```
jwmueller, line 77:
```suggestion
        If no images are found in the `data_path`
```

src/cleanvision/imagelab.py
```diff
@@ -59,13 +115,18 @@ def _set_default_config(self) -> Dict[str, Any]:
         }
 
     def list_default_issue_types(self) -> None:
+        """Prints a list of all issue types checked by default if no issue types are specified in imagelab.find_issues()"""
```
jwmueller, line 121:
```suggestion
        """Prints list of the issue types detected by default if no types are specified in ``imagelab.find_issues()``"""
```

src/cleanvision/imagelab.py
```diff
@@ -59,13 +115,18 @@ def _set_default_config(self) -> Dict[str, Any]:
...
         print("Default issue type checked by Imagelab:\n")
         print(
-            *[issue_type.value for issue_type in self.config["default_issue_types"]],
+            *[issue_type.value for issue_type in self._config["default_issue_types"]],
             sep="\n",
         )
 
     def list_possible_issue_types(self) -> None:
+        """Prints a list of all possible issue types that can be checked in the dataset.
+        It will also include custom added issue types.
```
jwmueller, line 131:
```suggestion
        """Prints list of all possible issue types that can be detected in a dataset.
        This list will also include custom issue types if you properly add them.
```

src/cleanvision/imagelab.py
```diff
@@ -87,15 +148,54 @@ def _get_issues_to_compute(
             }
         return to_compute_issues_with_params
 
-    def find_issues(self, issue_types: Optional[Dict[str, Any]] = None) -> None:
+    def find_issues(
+        self, issue_types: Optional[Dict[str, Dict[str, Any]]] = None
+    ) -> None:
+        """Finds issues in the dataset.
+        If issue_types is empty or not given, dataset is checked for all default issue types.
```
jwmueller, line 158:
```suggestion
        If `issue_types` is not provided, dataset is checked for a default set of issue types.
        To see default set: ``imagelabl.list_default_issue_types()``
```

src/cleanvision/imagelab.py
```diff
@@ -87,15 +148,54 @@ def _get_issues_to_compute(
...
+            issue_types = {
+                "dark": {"threshold": 0.1},
+                "blurry": {}
+            }
+            imagelab.find_issues(issue_types)
+
+        Different issue types can have different hyperparameters
+
+        """
         to_compute_issues_with_params = self._get_issues_to_compute(issue_types)
```
jwmueller, line 194: If `issue_types` is empty list, why do we check for default issue types? I think we should only check for default issue types when this is None. When it is empty list, should either raise ValueError or print("No issue types specified to detect") or something like that.

@elisno @sanjanag thoughts?

src/cleanvision/imagelab.py
```diff
@@ -211,6 +314,42 @@ def report(
         examples_per_issue: Optional[int] = None,
         verbosity: int = 1,
     ) -> None:
+        """Prints a summary of issues found in the dataset with their example images from the dataset.
+        By default, if no arguments are specified, it reports the top issues found in the dataset.
```
jwmueller, line 321:
```suggestion
        """Prints summary of the issues found in your dataset.
        By default, this method depicts the images representing top-most severe instances of each issue type.
```

src/cleanvision/imagelab.py
```diff
@@ -211,6 +314,42 @@ def report(
...
+
+        Parameters
+        ----------
+        issue_types : List[str], optional
+            List of issue types to report
```
jwmueller, line 326:
```suggestion
            List of issue types to consider in report.
            This must be subset of the issue types specified in ``imagelab.find_issues()``.
```

src/cleanvision/imagelab.py
```diff
@@ -211,6 +314,42 @@ def report(
...
+
+        num_top_issues : int, default=3
+            Number of top issues to report. It is ignored if issue_types is given
```
jwmueller, line 329:
```suggestion
            By default, this report will only cover the top most prevalent types of issues in the dataset, specified by this value. You can specify `issue_types` instead to override this and include specific types of issues in the report. 
```

src/cleanvision/imagelab.py
```diff
@@ -211,6 +314,42 @@ def report(
...
     ) -> None:
+        """Prints a summary of issues found in the dataset with their example images from the dataset.
+        By default, if no arguments are specified, it reports the top issues found in the dataset.
+
+        Parameters
+        ----------
+        issue_types : List[str], optional
+            List of issue types to report
+
+        num_top_issues : int, default=3
```
jwmueller, line 328: Should the default value be much larger, like 10?
I think somebody can easily understand 5-10 types of issues. Isn't it important to make somebody aware of 5 issues if they are all pretty severe in their dataset?

@elisno

src/cleanvision/imagelab.py
```diff
@@ -211,6 +314,42 @@ def report(
...
+            Number of top issues to report. It is ignored if issue_types is given
+
+        max_prevalence : float, default=0.5
+            Between 0 and 1.
+            Ignores the issue types from reporting if found in more than max_prevalence fraction of the dataset.
+            It is ignored if issue_types is given.
```
jwmueller, line 334:
```suggestion
            Value between 0 and 1, ignored if `issue_types` is provided. 
            Otherwise issue types that are detected in more than `max_prevalence` fraction of the images in dataset will be omitted from report.
            You are presumably already aware of these in your dataset.
```

src/cleanvision/imagelab.py
```diff
@@ -211,6 +314,42 @@ def report(
...
+
+        examples_per_issue : int, default=4
+            Number of examples to show for issue type reported.
```
jwmueller, line 337:
```suggestion
            Number of images to show for issue type reported. These are examples of the top-most severe instances of the issue in your dataset.
```

src/cleanvision/imagelab.py
```diff
@@ -211,6 +314,42 @@ def report(
...
+
+        verbosity : int, {1, 2, 3, 4}
+            Increasing verbosity increases the detail in the report output
```
jwmueller, line 340:
```suggestion
            Increasing verbosity increases the detail of the report. Set this to 0 to report less information, or to 4 to report the most information.
```

src/cleanvision/imagelab.py
```diff
@@ -301,17 +440,73 @@ def visualize(
...
+        """Visualization helper for images.
+
+        This method is multipurpose and can be used for visualizaing:
+        1. Random images from the dataset
+        2. Particular images with paths given in image_files
+        3. Top examples of given issue_types found in the dataset
+
+        If no image_files or issue_types are given, random images will be shown from the dataset.
+        If specific image_files are given, it will override the argument issue_types and will show given image_files
+        If image_files are not given and issue_types are given, top examples of given image_types will be shown
```
jwmueller, line 455:
```suggestion
        """Show specific images.
        
        Can be used for visualizing either:
        1. Particular images with paths given in `image_files`.
        2. Images representing top-most severe instances of given `issue_types` detected the dataset.
        3. If no `image_files` or `issue_types` are given, random images will be shown from the dataset.
        
        If `image_files` is given, this overrides the argument `issue_types`.
```

src/cleanvision/imagelab.py
```diff
@@ -301,17 +440,73 @@ def visualize(
...
+
+        Parameters
+        ----------
+
+        image_files : List[str], optional
+            List of image filepaths to visualize
```
jwmueller, line 461:
```suggestion
            List of filepaths for images to visualize.
```

src/cleanvision/imagelab.py
```diff
@@ -301,17 +440,73 @@ def visualize(
...
+
+        issue_types: List[str], optional
+            List of issue types to visualize
```
jwmueller, line 464:
```suggestion
            List of issue types to visualize. For each type of issue, will show a few images representing the top-most severe instances of this issue in the dataset.
```

src/cleanvision/imagelab.py
```diff
@@ -301,17 +440,73 @@ def visualize(
...
+
+        num_images : int, optional
+            Number of images to randomly visualize from the dataset
+            Used only when image_files and issue_types are empty, otherwise this argument will be ignored
+
+        examples_per_issue : int, optional
+            Number of top examples per issue type to visualize
+            Used only when issue_types is given and image_files is empty
```
jwmueller, line 472: what do you think about merging these two arguments to just be one argument: `num_images` ?

To me they have roughly the same effect (control how many images are plotted). In random case, it is the total number of images, otherwise it is the number of images per issue type.

src/cleanvision/imagelab.py
```diff
@@ -301,17 +440,73 @@ def visualize(
...
+
+        cell_size : Tuple[int, int]
+            cell size of each image in the image grid when visualizing
```
jwmueller, line 475:
```suggestion
            Dimensions controlling the size of each image in the depicted image grid.
```

src/cleanvision/imagelab.py
```diff
@@ -301,17 +440,73 @@ def visualize(
...
+
+        num_images : int, optional
+            Number of images to randomly visualize from the dataset
+            Used only when image_files and issue_types are empty, otherwise this argument will be ignored
+
+        examples_per_issue : int, optional
+            Number of top examples per issue type to visualize
+            Used only when issue_types is given and image_files is empty
+
+        cell_size : Tuple[int, int]
```
jwmueller, line 474:
```suggestion
        cell_size : Tuple[int, int], optional
```

src/cleanvision/imagelab.py
```diff
@@ -320,9 +515,14 @@ def get_stats(self) -> Any:
...
     def save(self, path: str) -> None:
-        """Saves this ImageLab to file (all files are in folder at path/).
+        """Saves this ImageLab instance at the given path.
         Your saved Imagelab should be loaded from the same version of the CleanVision package.
         This method does not save your image files.
+
+        Parameters
+        ----------
+        path : str
+            path at which to save the Imagelab instance
```
jwmueller, line 528:
```suggestion
            Path to folder where this Imagelab instance will be saved on disk.
```

src/cleanvision/imagelab.py
```diff
@@ -320,9 +515,14 @@ def get_stats(self) -> Any:
         return self.info["statistics"]
 
     def save(self, path: str) -> None:
```
jwmueller, line 520: overwriting an existing folder could be dangerous if somebody misspecifies the path.
What do you think about adding another argument:
`allow_overwrite = False` by default. 
If it's false, then this method raises ValueError if the path already exists telling user to set this argument to True.

What do you think? @elisno @sanjanag

src/cleanvision/imagelab.py
```diff
@@ -345,11 +545,22 @@ def save(self, path: str) -> None:
...
-        `data_path` is the path to image dataset previously used in Imagelab.
-        If the `data_path` is changed, Imagelab will not be loaded as some of its functionalities depend on it.
-        You should be using the same version of the CleanVision package previously used when saving the Imagelab.
+        """Loads Imagelab from given path.
+
+
+        Parameters
+        ----------
+        path : str
+            Path to the saved Imagelab, not pickle file.
```
jwmueller, line 557:
```suggestion
            Path to the saved Imagelab folder previously specified in ``save()`` (not the individual pickle file).
```

src/cleanvision/imagelab.py
```diff
@@ -345,11 +545,22 @@ def save(self, path: str) -> None:
...
+        data_path : str
+            Path to image dataset previously used in Imagelab.
+            If the data_path is changed, Imagelab will not be loaded as some of its functionalities depend on it.
```
jwmueller, line 560:
```suggestion
            If the `data_path` is changed, Imagelab will not be loaded as some of its functionalities depend on it.
```

src/cleanvision/imagelab.py
```diff
@@ -1,3 +1,9 @@
+"""
```
jwmueller, line 1: High-level feedback: Use numpy format for docstrings.
Argument names should be enclosed in single backticks: `x`.
Code snippets should be enclosed in double backticks: ``method(...)``. 

When referencing classes/methods that are part of CleanVision, it would be nice to link to them. How to do this is described here:
https://github.com/cleanlab/cleanlab/blob/master/DEVELOPMENT.md#documentation-style

src/cleanvision/utils/utils.py
sanjanag: I replaced this because I was editing while going through the numpy docstyle page and one of the examples was like this. But in other parts I stuck to other type of annotation and didn't come back to this.

src/cleanvision/imagelab.py
sanjanag: Checking for default issues only when None, sounds the right behaviour to me for `find_issues`

src/cleanvision/imagelab.py
sanjanag: Yeah, I can merge them.

src/cleanvision/imagelab.py
sanjanag: Added

src/cleanvision/imagelab.py
sanjanag: Renamed

src/cleanvision/imagelab.py
sanjanag: It's a good argument to have. I've added.

src/cleanvision/imagelab.py
jwmueller: @sanjanag Can we raise ValueError if both data_path and file_paths are provided?  And if they are both None

Seems like user messed something up in that case.

src/cleanvision/imagelab.py
jwmueller: @sanjanag I don't think this has been addressed?

src/cleanvision/imagelab.py
jwmueller: @sanjanag I don't think this has been addressed?

src/cleanvision/imagelab.py
```diff
@@ -303,41 +478,116 @@ def visualize(
...
+        image_files : List[str], optional
+            List of filepaths for images to visualize.
+
+        issue_types: List[str], optional
+            List of issue types to visualize. For each type of issue, will show a few images representing the top-most severe instances of this issue in the dataset.
+
+        num_images : int, optional
+            Number of images to randomly visualize from the dataset.
+            If `issue_types` is given, it implies number of top examples per issue type to visualize.
+            If `image_files` is given, this argument will be ignored.
```
jwmueller, line 504:
```suggestion
            Number of images to visualize from the dataset.
            These are randomly selected if `issue_types` is ``None``.
            If `issue_types` is given, then this is the number of images for each issue type to visualize
            (images representing top-most severe instances of this issue will be shown).
            If `image_files` is given, this argument is ignored.
```

src/cleanvision/imagelab.py
```diff
@@ -90,14 +182,55 @@ def _get_issues_to_compute(
     def find_issues(
         self, issue_types: Optional[Dict[str, Any]] = None, n_jobs: Optional[int] = None
     ) -> None:
+        """Finds issues in the dataset.
+        If `issue_types` is not provided, dataset is checked for a default set of issue types.
+        To see default set: :py:meth:`Imagelab.list_default_issue_types`
+
+        Parameters
```
jwmueller, line 189: `n_jobs` needs documentation. I recommend following documentation from these places:

https://docs.cleanlab.ai/master/cleanlab/filter.html#cleanlab.filter.find_label_issues

> Number of processing threads used by multiprocessing. Default None sets to the number of cores on your CPU (physical cores if you have psutil package installed, otherwise logical cores). Set this to 1 to disable parallel processing (if it's causing issues).

src/cleanvision/imagelab.py
```diff
@@ -303,41 +483,118 @@ def visualize(
...
+        """Saves this ImageLab instance into a folder at the given path.
         Your saved Imagelab should be loaded from the same version of the CleanVision package.
         This method does not save your image files.
+
+        Parameters
+        ----------
+        path : str
+            Path to folder where this Imagelab instance will be saved on disk.
+        allow_overwrite: bool, default=False
+            If set to True, existing files at `path` will be overwritten.
```
jwmueller, line 576:
```suggestion
        path : str
            Path to folder where this Imagelab instance will be saved on disk.

        allow_overwrite: bool, default=False
            If set to True, any existing files at `path` will be overwritten.
```

src/cleanvision/imagelab.py
```diff
@@ -303,41 +483,118 @@ def visualize(
...
+        if not path_exists:
             os.mkdir(path)
+        else:
+            if allow_overwrite:
+                print(
+                    f"WARNING: Existing files will be overwritten by newly saved files at: {path}"
+                )
+            else:
+                raise ValueError(
+                    "Path already exists, specify a new path or set allow_overwrite=True"
```
jwmueller, line 593:
```suggestion
                    "Existing files already exist at the given path, either specify a new path or set allow_overwrite=True"
```

jwmueller (approved):
had two more nit suggestions but otherwise LGTM.

@elisno please look over the comments where you are tagged as these decisions influence Datalab

src/cleanvision/imagelab.py
elisno: This was already the case in Datalab in one of the internal methods that sets necessary arguments based on those conditions.

I've added an early return for this comment, but that's not covered by tests.

src/cleanvision/imagelab.py
elisno: If we're going to use an `allow_overwrite`, then we need to raise a [FileExistsError](https://docs.python.org/3/library/exceptions.html#FileExistsError).

src/cleanvision/imagelab.py
jwmueller: I don't think it is worth switching to class yet, we don't want to delay anymore on this project. I've added github issue [here](https://github.com/cleanlab/cleanvision/issues/102) where @elisno can provide more pointers and the discussion can continue.

jwmueller: > Should we have a (+123 other types of issues) line at the end of the report if we still worry about users "missing" severe issue types?

Not in my opinion. The things that are over max_prevalence are not issues. If the dataset is all black/white, or all blurry, these are basically just not issues in the dataset that need to be reported to user. User can still easily figure these out, by looking at Imagelab attributes (or just browsing their data), but these are very likely low-priority for the user. 

In Imagelab & Datalab reports, we **always** want to err on the side of showing **less** not more. The purpose of these is to focus attention where it is needed most.

sanjanag: Changed the `imagelab.report()`
- Removed num_top_issues as it was a useless argument, number of issues output can be increased by verbosity.
- For `max_prevalence`, first filter out issues in both `issue_types` argument or if no issue types are provided according to max_prevalence and then report the issues.
- Added a print statement to indicate if there were more issues found in the dataset and how to view them.

PR #93, opened 2023-02-09
Updated dev guide

sanjanag:
- Updated a few commands
- Removed $ sign from the start of commands to avoid it being copied with the command


jwmueller (approved):
good call!

PR #99, opened 2023-02-16
Bug fixes

sanjanag:
- Replaced ImageGrid visualization with a custom grid created using `plt.subplots()` as ImageGrid was cropping images to fit into axes
- Added n_jobs in type check for issue manager

----
:books: Documentation preview :books:: https://cleanvision--99.org.readthedocs.build/en/99/

### DIFF
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -277,9 +277,6 @@ class Imagelab:
                     filepaths=sorted_filepaths,
                     ncols=self.config["visualize_num_images_per_row"],
                     cell_size=cell_size,
-                    cmap="gray"
-                    if issue_type_str == IssueType.GRAYSCALE.value
-                    else None,
                 )
         elif viz_name == "image_sets":
             image_sets = self.info[issue_type_str][SETS][:examples_per_issue]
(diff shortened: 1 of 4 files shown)


src/cleanvision/imagelab.py
```diff
@@ -277,9 +277,6 @@ def _visualize(
                     filepaths=sorted_filepaths,
                     ncols=self.config["visualize_num_images_per_row"],
                     cell_size=cell_size,
```
jwmueller, line 279: have you checked the code works if some images are grayscale, eg. running CleanVision on MNIST?

jwmueller (approved):
FYI type check complains:

> src/cleanvision/issue_managers/duplicate_issue_manager.py:137: error: Returning Any from function declared to return "Union[SupportsDunderLT[Any], SupportsDunderGT[Any]]"  [no-any-return]

If not easily resolvable, may be worth telling mypy to ignore this part and opening github issue to revisit this in the future

src/cleanvision/imagelab.py
sanjanag: Yes, it works. The check is now in this [line](https://github.com/cleanlab/cleanvision/pull/99/files#diff-6c1df1b10a098d41817df84dea378a7e519e90791ecdb2baa738bdbdba027b2eR27)

PR #103, opened 2023-02-18
limit max number of processes for multiprocessing

clu0:
if the number of images is too large (e.g. > 10M), previous implementation with multiprocessing could become slow.

Limits max number of processes to `MAX_PROCS=5000`

----
:books: Documentation preview :books:: https://cleanvision--103.org.readthedocs.build/en/103/


src/cleanvision/utils/utils.py
```diff
@@ -12,6 +12,9 @@
 except ImportError:  # pragma: no cover
     PSUTIL_EXISTS = False
 
+# max number of processes that can be forked/spawned for multiprocessing
+MAX_PROCS = 5000
```
sanjanag, line 16: move this to `utils/constants.py`

sanjanag: which dataset did you test it on?

clu0: yup moved to `utils/constants.py`.

I didn't actually observe the slowdown for this repo, because I didn't have any dataset with millions of images to test on. I made this change because when doing multiprocessing for `find_label_issues` in the cleanlab repo, we observed that when the number of processes gets into the millions suddenly the code becomes slow. So this is more a preventative change, I don't think it hurts.

sanjanag: Okay cool

sanjanag (approved):

PR #105, opened 2023-02-22
Phash

sanjanag:
Bug fixes:
- Visualize only top issues in `imagelab.report()`
- Raise exception on passing an empty list of files in `imagelab.visualize()`

Improvements:
- Updated thresholds for issue types based on these datasets : our example dataset, cifar10, caltech256, food101, cubs2011, weather, intel
- Changed logic of dark and light issues: get nth percentile for. brightness of each pixel tune that value for threshold, helps with false positives containing objects in dark/light backgrounds.
- Show scores on top of each image instead of filename in `imagelab.visualize(issue_types)`
- Changed default hash near_duplicates hash to `phash`
- For exact duplicates calculate pixel value hash instead of image object hash

Refactoring
- ImageProperty _scores_ passed around in the code as pandas series instead of list, also helps in maintaining indexes.
- `ImageProperty.calculate` returns dict instead of a single value
- Introduced `score_column` property in ImageProperty for specifying which series to use for evaluating scores

----
:books: Documentation preview :books:: https://cleanvision--105.org.readthedocs.build/en/105/

### DIFF
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -431,9 +431,9 @@ class Imagelab:
 
         self.visualize(
             issue_types=filtered_issue_types[:num_report],
-            num_images=report_args["examples_per_issue"]
-            if num_images is None
-            else num_images,
+            num_images=(
+                report_args["examples_per_issue"] if num_images is None else num_images
+            ),
         )
 
     def _pprint_issue_summary(self, issue_summary: pd.DataFrame) -> None:
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -28,6 +28,7 @@ from cleanvision.utils.utils import (
     get_filepaths,
     deep_update_dict,
     get_is_issue_colname,
+    get_score_colname,
 )
 from cleanvision.utils.viz_manager import VizManager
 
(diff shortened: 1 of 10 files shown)


src/cleanvision/issue_managers/image_property_issue_manager.py
```diff
@@ -162,49 +158,100 @@ def find_issues(
...
+                    else:
+                        self.info[issue_type][column_name] = comp[column_name]
+            else:
+                if property_name not in self.info["statistics"]:
+                    self.info["statistics"][property_name] = imagelab_info[
+                        "statistics"
+                    ][property_name]
+                self.info[issue_type] = imagelab_info.get(issue_type, {})
+
+        # todo: revisit when there are more issues using same properties like light and dark
```
elisno, line 237: Is the intention to map compatible issues to each other based on the shared property? 

Suppose an issue (e.g. LIGHT) is valid (exists in `self.issue_types`) but is not set in `self.info`. Do we assume its counterpart (e.g. DARK) has been set with the desired property?

tests/test_image_property_helpers.py
```diff
@@ -44,47 +44,48 @@ def test_get_image_mode(image, expected_mode):
...
+    #     raw_scores = [0.5, 0.3, 1.0, 1.2, 0.9, 0.1, 0.2]
+    #     expected_output = np.array([0.5, 0.3, 1.0, 1.0, 0.9, 0.1, 0.2])
+    #
+    #     with monkeypatch.context() as m:
+    #         m.setattr(image_property, "issue_type", IssueType.DARK)
+    #         normalized_scores = image_property.get_scores(raw_scores=raw_scores)
+    #         assert all(normalized_scores==expected_output)
+    #
+    #     normalized_scores = image_property.get_scores(raw_scores=raw_scores)
+    #     assert all(normalized_scores==1 - expected_output)
```
elisno, line 88: Todo comment is fine.

Instead of commenting out the tests, how about decorating them with something like `skip` or `xfail` markers?

https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-test-functions
https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

jwmueller (approved):
messaged you all my comments for easier discussion, but I think they can be addressed in future PRs (or this one if you can get to them in time and then re-request my review)

src/cleanvision/imagelab.py
```diff
@@ -449,43 +452,63 @@ def _get_issue_manager(self, issue_type_str: str) -> IssueManager:
...
-                    f"\nTop {examples_per_issue} sets of images with {issue_type_str} issue"
+                    f"\nTop {num_images} {sets_str} of images with {issue_type} issue"
                 )
+
+            title_sets = []
+            for s in image_sets:
+                filenames = []
+                for path in s:
+                    filenames.append(path.split("/")[-1])
+                title_sets.append(filenames)
```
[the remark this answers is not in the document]
sanjanag: Yes

elisno (approved):

src/cleanvision/issue_managers/image_property_issue_manager.py
sanjanag: Yes. In `imagelab.find_issues` I get a defer set of issues to re-use computations. This defer set will add, all but one of the compatible issues to the defer set. These left out issues will be populated using the one that was left out of defer set.

PR #107, opened 2023-02-27
Phash new

sanjanag:
- Changes from old `phash` branch
- Added a couple of bug fixes
- added jupyter notebook again after removing outputs

----
:books: Documentation preview :books:: https://cleanvision--107.org.readthedocs.build/en/107/

### DIFF
--- a/examples/demo.ipynb
+++ b/examples/demo.ipynb
@@ -11,13 +11,6 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Installation"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
     "For installing cleanvision, you will have to first clone the repo and then install from source.\n",
     "Steps for installing from cloned repo\n",
     "```bash\n",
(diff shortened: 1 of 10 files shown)


examples/demo.ipynb
```diff
@@ -0,0 +1,560 @@
...
+   "metadata": {},
+   "source": [
+    "# Getting Started"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Installation"
```
jwmueller, line 14: IMO you can get rid of this subheader to save space,  since it's obvious what this section is about

examples/demo.ipynb
```diff
@@ -0,0 +1,560 @@
...
+   ]
```
jwmueller, line 15: instead it would be nice to add a colab badge here which opens the notebook in colab. Assume it will be opened from the main branch, as you can see is done in our examples notebooks here: 
https://github.com/cleanlab/examples/

jwmueller (approved):
seems good to me at a high level, I cannot closely inspect every detail though

examples/demo.ipynb
sanjanag: I'll add it in the next PR since to actually make the link clickable, need the notebook on main branch.

PR #108, opened 2023-02-27
Increase codecov

ulya-tkch:
----
:books: Documentation preview :books:: https://cleanvision--108.org.readthedocs.build/en/108/


tests/test_run.py
```diff
@@ -0,0 +1,170 @@
...
+    # Check for additional types of issues using existing Imagelab
+    imagelab = Imagelab.load(save_folder, generate_n_image_files)
+    issue_types = {"light": {}, "low_information": {}}
+    imagelab.find_issues(issue_types)
+    imagelab.report()
+
+    # Check for an issue with a different threshold
+    issue_types = {"dark": {"threshold": 0.2}}
+    imagelab.find_issues(issue_types)
+    imagelab.report(issue_types=issue_types.keys())  # report only specific issues
```
sanjanag, line 132: are you planning to add any assert statements here?

ulya-tkch: not in this iteration just trying to get the code covered

tests/test_duplicate_issue_manager.py
```diff
@@ -198,3 +202,23 @@ def test_get_duplicate_sets(
...
+    assert hash is not None
+
+    hash = get_hash(img, {"hash_type": "whash", "hash_size": 2})
+    assert hash is not None
+
+    hash = get_hash(img, {"hash_type": "phash", "hash_size": 2})
+    assert hash is not None
+
+    # Test calling function with not a real hash
+    try:
```
sanjanag, line 219: https://www.educative.io/answers/how-to-check-if-an-exception-gets-raised-in-pytest
Check example 2
`import pytest

def func(x):
  if x == 5:
    raise ValueError("x must be a value other than 5")
  return x

def test_func():
  with pytest.raises(ValueError, match="x must be a value other than 5"):
    func(5)

# func(5)     # Raises an exemption

test_func()`

sanjanag (approved):

PR #116, opened 2023-03-10
Readme updates

sanjanag:


README.md
```diff
@@ -5,19 +5,42 @@
...
+project to find problems in your dataset, which you may want to address before applying machine learning. It currently detects issues in the raw images themselves, and can be useful for any machine learning task (classification, segmentation, object detection, pose estimation, keypoint detection, image generation, etc).
+The list of all issues detected by CleanVision can be found in the table below.
+
+
+|   | Issue Type       | Description                                               | Issue Key        | Example                                      |
+|---|------------------|-----------------------------------------------------------|------------------|----------------------------------------------|
+| 1 | Dark             | Irregularly dark images                                   | dark             | ![](docs/readme_images/dark.jpg)             |
+| 2 | Blurry           | Blurry or out of focus images                             | blurry           | ![](docs/readme_images/blurry.png)           |
+| 3 | Grayscale        | Images lacking color                                      | grayscale        | ![](docs/readme_images/grayscale.png)        |
```
jwmueller, line 16: I would only show the left image out of these two, is there a reason to show 2 images for "grayscale"? 

Also if possible, I'd choose an image that is more obviously black&white rather than these images where it isn't 100% obvious.

Screen Shot 2023-03-09 at 10 52 42 PM

README.md
```diff
@@ -5,19 +5,42 @@
 
 CleanVision automatically detects various issues in image datasets, such as images that are: (near) duplicates,
 blurry, over/under-exposed, etc. This data-centric AI package is designed as a quick first step for any computer vision
-project to find problems in your dataset, which you may want to address before applying machine learning.
+project to find problems in your dataset, which you may want to address before applying machine learning. It currently detects issues in the raw images themselves, and can be useful for any machine learning task (classification, segmentation, object detection, pose estimation, keypoint detection, image generation, etc).
+The list of all issues detected by CleanVision can be found in the table below.
```
jwmueller, line 9: IMO this extra text makes this no longer a nice short paragraph description of the library.

I would move all the extra text "It currently detects ..." later in the readme, as this is starting to get into details that do not need to appear in the TLDR description of the library.

README.md
```diff
@@ -5,19 +5,42 @@
```
jwmueller, line 5: I think the goal of the readme is to present the interesting highlights / most useful functionalities first, before listing everything the package can do. Thus I think optimal organization of the readme (which is closer to what readme looked like before this restructuring) would be:

- CleanVision title
- Banner image (that @ulya-tkch is working on, illustrating the purpose of the library)
- short paragraph description of the library
- quickstart code example
- more links on how to get started
- badges 
- More description of library, eg: your the extra sentences "It currently detects issues in the raw images ..." 
- Table of issue types

@sanjanag What do you think? Why did you prefer to restructure the library as you did in this PR?

README.md
```diff
@@ -5,19 +5,42 @@
 
 CleanVision automatically detects various issues in image datasets, such as images that are: (near) duplicates,
 blurry, over/under-exposed, etc. This data-centric AI package is designed as a quick first step for any computer vision
-project to find problems in your dataset, which you may want to address before applying machine learning.
+project to find problems in your dataset, which you may want to address before applying machine learning. It currently detects issues in the raw images themselves, and can be useful for any machine learning task (classification, segmentation, object detection, pose estimation, keypoint detection, image generation, etc).
+The list of all issues detected by CleanVision can be found in the table below.
+
+
+|   | Issue Type       | Description                                               | Issue Key        | Example                                      |
```
jwmueller, line 12: Consider making all images in this table the same height for a more polished look

README.md
```diff
@@ -26,53 +49,28 @@ imagelab.find_issues()
 imagelab.report()
 ```
 
-CleanVision diagnoses many types of issues, but you can also check for only specific issues:
-
+3. This step is optional. You can also specify specific issues instead of checking for all issues.
+The **Issue Key** column in the table above specifies the name for each type of issue in the CleanVision code. Use these string names to specify which types of issues should be checked.
```
jwmueller, line 53: IMO this is too much details for a readme quickstart. So I'd just keep the old text:

> CleanVision diagnoses many types of issues, but you can also check for only specific issues:

I think it's fairly obvious how to check for specific types of issues based on the code itself

README.md
```diff
@@ -26,53 +49,28 @@ imagelab.find_issues()
...
 ```
 
-More on how to get started with CleanVision:
-
-- [Example script](https://github.com/cleanlab/cleanvision/blob/main/examples/run.py)
-- [Jupyter notebook tutorial](https://github.com/cleanlab/cleanvision/blob/main/examples/demo.ipynb)
+## More on how to use CleanVision
+- [Tutorial notebook](https://github.com/cleanlab/cleanvision/blob/main/examples/demo.ipynb)
 - [Documentation](https://cleanvision.readthedocs.io/)
+- [Example script](https://github.com/cleanlab/cleanvision/blob/main/examples/run.py)
```
jwmueller, line 67: link to the cleanvision-examples repo here too?
https://github.com/cleanlab/cleanvision-examples

jwmueller (changes requested):
I'm not sure the new overall readme structure is preferable to how it was before, so curious to hear your reasoning. But the added images to the table are great!

README.md
sanjanag: The intention in moving the table up was to highlight all the issues that the package can detect for someone new. But I see your point how it occupies a lot of important space in the starting of readme I checked some popular repos on github and I propose these sections
- CleanVision title (everyone has a title image, but we can skip it since it's ultimately going to be merged and don't want a separate identity)
- badges (This doesn't occupy much space but makes it look more professional)
- Banner image
- Overview (description of what cleanvision does)
- Installation (while it is not a big section, I think it makes sense to have this separately from a user perspective)
- Quickstart (code example)
- Table of issue types
- Resources (tutorial notebook, example notebooks, documentation, example script, example dataset)
- Remaining sections (like contributing)

jwmueller: I would push the Table of issue-types under "Getting started" resources which has critical links to docs, examples, etc.

In the future, we can have a second longer "Resources" section with miscellaneous links that goes under the Table of issue types.

jwmueller: I would put the badges under the overview paragraph, they basically demarcate the above-the-fold content from the rest IMO

jwmueller: I think installation can just be part of quickstart as you have it now.

jwmueller: between Quickstart and Table of issue types, there should be more in-depth explanations about the library can be useful for, eg. your sentences starting with: "It currently detects issues in the raw images ..."

jwmueller: FYI beware about thinking badges always make repos look more professional, the CI badge can easily say "failing", etc. which will not contribute to a professional look

README.md
sanjanag: I resized all images to same height but the near and exact duplicate images get resized when rendering and look comparitively smaller.

jwmueller: ah no worries about that

jwmueller (approved):
made a few more edits, please look them over and then feel free to merge or edit more if you still have further improvements!

jwmueller:
Actually wait before you merge, please move all image files out of this PR and into the assets repo (like we did for the cleanvision banner image). We do not want to clutter this repo with image files.

PR #125, opened 2023-03-15
Docs changes

sanjanag:
Addresses these comments
https://cleanlabinc.slack.com/archives/C04DC8K716X/p1678837871858979

### DIFF
--- a/src/cleanvision/imagelab.py
+++ b/src/cleanvision/imagelab.py
@@ -589,7 +589,7 @@ class Imagelab:
 
     # Todo: Improve mypy dict typechecking so this does not return any
     def get_stats(self) -> Any:
-        """Returns statistics on images like brightness, color space, aspect ratio"""
+        """Returns dict of statistics computed from images when auditing the data such as: brightness, color space, aspect ratio, etc."""
         return self.info["statistics"]
 
     def save(self, path: str, force: bool = False) -> None:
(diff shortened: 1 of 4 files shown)


src/cleanvision/imagelab.py
```diff
@@ -589,6 +589,7 @@ def visualize(
 
     # Todo: Improve mypy dict typechecking so this does not return any
     def get_stats(self) -> Any:
+        """Returns statistics on images like brightness, color space, aspect ratio"""
```
jwmueller, line 592:
```suggestion
        """Returns statistics computed about images when auditing the data such as: brightness, color space, aspect ratio, etc."""
```

jwmueller (approved):
just had one suggestion

src/cleanvision/imagelab.py
jwmueller: say "dict of statistics" instead of "statistics" if this indeed returns a dict

jwmueller: In future would be good to clarify when methods return `None`, I assume this one does if the stats haven't been computed yet

sanjanag: Yes, this can return None.

ulya-tkch (approved):
LGTM!

src/cleanvision/imagelab.py
jwmueller: I was saying you should clarify this in the docstring. Adding it myself now.

PR #126, opened 2023-03-16
Improved writing

sanjanag:


examples/demo.ipynb
```diff
@@ -30,7 +30,7 @@
    "cell_type": "markdown",
```
jwmueller, line 30: Let's hold off updating this notebook until the corresponding version in `cleanvision-examples/` is merged. Then you can just clear the outputs of that one and merge it here (no review needed)

jwmueller (changes requested):
Let's hold off updating this notebook until the corresponding version in cleanvision-examples/ is merged. Then you can just clear the outputs of that one and merge it here (no review needed)

jwmueller (approved):

PR #131, opened 2023-03-20
Added limits to prevent matplotlib from unusual normalization

sanjanag:


jwmueller (approved):
LGTM, just make sure you re-ran all the benchmarks/notebooks with this new setting

PR #142, opened 2023-03-29
Remove .JPEG & .PNG from extension list as .jpeg & .png are already present.

GirinChutia:
**BUG**
When a image folder has .jpeg or .png extension images. Imagelab loads that image twice as shown below,

image

Its happening because in  `src/cleanvision/utils/utils.py` the extension lists have both '.JPEG' and '.jpeg', same for .png as well.

image

When loaded in the downstream code (by looping the extension list) using glob module to load the images, if a image has '.JPEG' extension, it load once for the '.JPEG' and once again for '.jpeg' because glob module treats both '.JPEG' and '.jpeg' as same image i.e, it doesn't matters if the extension is all caps or not.

**Solution**
Just keep '.jpeg' and '.png' in the list. It will handle even if the extension of the image is in all caps.

### DIFF
--- a/src/cleanvision/utils/utils.py
+++ b/src/cleanvision/utils/utils.py
@@ -21,8 +21,6 @@ TYPES: List[str] = [
     "*.jp2",
     "*.TIFF",
     "*.WebP",
-    "*.PNG",
-    "*.JPEG",
     "*.png",
 ]  # filetypes supported by PIL
 


jwmueller:
Closing this PR in favor of: https://github.com/cleanlab/cleanvision/pull/143

Note the bug was just on Windows, so thanks a ton for pointing it out.  Feel free to make other PRs if you see more opportunities for improvement!

PR #156, opened 2023-04-12
Updated torch n_jobs=1

sanjanag:


src/cleanvision/issue_managers/image_property_issue_manager.py
```diff
@@ -145,6 +147,8 @@ def find_issues(
         if to_be_computed:
             if n_jobs is None:
                 n_jobs = get_max_n_jobs()
+            if isinstance(dataset, TorchDataset):
```
jwmueller, line 150: See my other comment about this only being good to do when n_jobs is None (make sure not to accidentally just check if it's none here since you are overwriting the none value in the earlier lines).


Just curious: why do we have to overwrite n_jobs twice? Isn't there only one call to find_issues() where we could override it once?

jwmueller (changes requested):
Had a couple questions/suggestions

examples/huggingface_dataset.ipynb
```diff
@@ -21,7 +21,7 @@
    "outputs": [],
    "source": [
     "!pip install -U pip\n",
-    "!pip install cleanvision[huggingface]"
+    "!pip install \"cleanvision[huggingface]\""
```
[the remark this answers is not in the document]
sanjanag: It worked on colab without the quotes, but it breaks on my local jupyter notebook.

src/cleanvision/issue_managers/image_property_issue_manager.py
sanjanag: Yeah, it's a good idea to move this n_jobs logic out in the imagelab.find_issues()

jwmueller (approved):
LGTM! Don't forget to update the cleanvision-examples notebooks too

Issue #159, opened 2023-04-15
Export current version as __version__


anishathalye:
cleanvision doesn't currently export a `cleanvision.__version__`, this would be nice to do.

PR #160, opened 2023-04-17
allow cleanvision.__version__ to work

jwmueller:
Closes https://github.com/cleanlab/cleanvision/issues/159

Not sure this is the optimal way to do it, so feel free to suggest alternative. I just followed the tips from here:
https://stackoverflow.com/questions/67085041/how-to-specify-version-in-only-one-place-when-using-pyproject-toml

### DIFF
--- a/src/cleanvision/__init__.py
+++ b/src/cleanvision/__init__.py
@@ -1,5 +1,6 @@
-try:  # not supported on Python 3.7
+try:
     import importlib.metadata
+
     __version__ = importlib.metadata.version("cleanvision")
 except Exception:
     pass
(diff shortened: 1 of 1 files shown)


sanjanag:
I tested this it looks good, it only works for python>=3.8 though, as before this package was named `importlib_metadata` instead of `importlib.metadata`

src/cleanvision/__init__.py
```diff
@@ -0,0 +1,6 @@
+try:
+    import importlib.metadata
+
```
sanjanag, line 3:
```suggestion

# import only works for python>=3.8
```

jwmueller: added in a different line

PR #162, opened 2023-04-19
Added worfklow for automatic release

sanjanag:
Added a release pipeline with 3 jobs that will be executed in order:
1. Publish to TestPypi
2. Publish to Pypi
3. Bump patch version of the package and commit

This workflow will be triggered whenever the `CI` workflow is triggered in `main` branch **or** a release is published. However for the action of uploading the package to TestPypi and Pypi, `CI` workflow should be successful on `main` branch **and** release should be published.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches: [ main ]
   pull_request:
+  release:
+    types: [ published ]
 
 jobs:
   test:
@@ -20,12 +22,6 @@ jobs:
         run: pip install build
       - name: Build distributions
         run: python -m build
-      - name: Publish package to Test PyPI
-        if: github.repository == 'cleanlab/cleanvision' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
-        uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-          repository-url: https://test.pypi.org/legacy/
       - name: Publish package to PyPI
         if: github.repository == 'cleanlab/cleanvision' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
         uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,6 @@
 name: Build and publish distribution
 
+# This workflow will be triggered whenever the CI workflow is triggered in main branch is completed or a release is published
 on:
   workflow_run:
     workflows: [ "CI" ]
@@ -24,13 +25,15 @@ jobs:
       - name: Build distributions
         run: python -m build
       - name: Publish package to PyPI
-        if: github.repository == 'cleanlab/cleanvision' && github.event.workflow_run.conclusion == 'success'
+        # This condition ensures the CI workflow in main branch was successful and this is indeed a release published action
+        if: github.repository == 'cleanlab/cleanvision' && github.event.workflow_run.conclusion == 'success' && github.event_name == 'release' && github.event.action == 'published'
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           password: ${{ secrets.TEST_PYPI_API_TOKEN }}
           repository-url: https://test.pypi.org/legacy/
   publish:
     needs: [ publish-test ]
+    # This condition ensures publish-test succeeded before running this job
     if: needs.publish-test.result == 'success'
     runs-on: "ubuntu-latest"
     steps:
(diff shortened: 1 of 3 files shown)


jwmueller (approved):
Looks good AFAICT

ulya-tkch (approved):
Publishing to PyPI lgtm. Do we want to consider publishing to Test PyPI as well?

```python3
- name: Publish package to TestPyPI
  uses: pypa/gh-action-pypi-publish@release/v1
  with:
    password: ${{ secrets.TEST_PYPI_API_TOKEN }}
    repository-url: https://test.pypi.org/legacy/
```

elisno (changes requested):

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
+name: Build distribution
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
```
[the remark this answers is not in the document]
sanjanag: The idea to keep it on pull_request and main push were to ensure that the package can be built, but it is published only on release action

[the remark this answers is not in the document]
elisno: That's fine, I see you're adding the Test PyPI again, where the push and PR events make more sense.

.github/workflows/release.yml
```diff
@@ -1,16 +1,17 @@
...
     branches: [ main ]
-  pull_request:
+    types:
+      - completed
   release:
     types: [ published ]
 
 jobs:
-  test:
+  publish:
```
elisno, line 13: How about you make a duplicate job in this file?

Call this one `publish-test`, the subsequent one `publish` then `needs: publish-test`.

([Docs link](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#example-requiring-successful-dependent-jobs))

`publish-test` publishes to Test PyPI, the latter to the actual PyPI repo.

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Bump version
+        run: bump2version --current-version "${{ env.CURRENT_VERSION }}" patch pyproject.toml
+      - name: Commit and push
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          file_pattern: pyproject.toml
+          branch: autobump
+          commit_message: Bump patch version
+
+
```
jwmueller, line 84: don't forget to add the instructions we discussed to Development.md about person who publishes the release needing to wait for unit tests to pass first

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+      - name: Install build dependencies
+        run: pip install build
+      - name: Build distributions
+        run: python -m build
+      - name: Publish package to PyPI
+        # This condition ensures the CI workflow in main branch was successful and this is indeed a release published action
+        if: github.repository == 'cleanlab/cleanvision' && github.event.workflow_run.conclusion == 'success' && github.event_name == 'release' && github.event.action == 'published'
```
elisno, line 29: Correct me if I'm wrong, @sanjanag, but aren't `workflow_run` and `release` disjoint events?

Is this the correct expression? (Double-check my placement of parentheses)

```suggestion
        if: github.repository == 'cleanlab/cleanvision' && ( github.event.workflow_run.conclusion == 'success' || ( github.event_name == 'release' && github.event.action == 'published' ) )
```

DEVELOPMENT.md
```diff
@@ -109,7 +109,15 @@ endings match the project style.
 
 ## Merging PRs to the codebase
 
-The docs for this repo are hosted by readthedocs.io and can be found [here](https://cleanvision.readthedocs.io/en/latest/). 
+The docs for this repo are hosted by readthedocs.io and can be found [here](https://cleanvision.readthedocs.io/en/latest/).
 Currently, merged PRs *will not* update the docs -- you **need** to do this manually after merging a PR.
```
ulya-tkch, line 113: Is this still true? I think PRs merged to main will update docs!

sanjanag: No

.github/workflows/release.yml
sanjanag: Yes, they are disjoint events. My reason for adding workflow_run to the condition was to ensure the `CI` workflow was successful in the `main` branch. With the OR condition, this will release on main in merge also, which I did not want, hence the AND condition.

elisno: I guess what is confusing me is: If the workflow_run condition is true, then the CI wasn't triggered by a release event, so the latter part of the expression will be false?

sanjanag: you are right, I can't place a check on disjoint events. Removed workflow_run since it was not serving the intended purpose.

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+      - name: Install build dependencies
+        run: pip install build
+      - name: Build distributions
+        run: python -m build
+      - name: Publish package to PyPI
+        if: github.repository == 'cleanlab/cleanvision' && github.event_name == 'release' && github.event.action == 'published'
```
elisno, line 44: This is no longer necessary, right?

You have this single trigger event for the workflow
```yml
on:
  release:
    types: [ published ]
```


along with the `if: needs.PREVIOUS_JOB.result == 'success'` check.

elisno (approved):
LGTM!

PR #164, opened 2023-04-22
Incremented version

sanjanag:


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "cleanvision"
-version = "0.2.1"
+version = "0.2.2"
```
jwmueller, line 7: should be 0.2.3 so you can make pypi work?

jwmueller (approved):

PR #166, opened 2023-04-26
Datalab integration

sanjanag:
- Changed signature of `list_default_issue_types` and `list_possible_issue_types` to return `str` instead of printing output
- Added `verbose` arg to `imagelab.find_issues(verbose=True)` to suppress cleanvision logging in finding issues when called from datalab
- Removed control of number of images shown per issue type in `imagelav.report()` from report's `verbosity` argument, as this can be controlled by `num_images` and didn't seem the right usage of verbosity.
- Added `print_summary` arg to `imagelab.report(print_summary=True)` to suppress printing a separate report in Datalab
- Updated tests for the package to be more meaningful and extensive.
- For functional testing, also added a small dataset that covers all kinds of issues.

- [x] Add documentation for new args


elisno (approved):
LGTM!

jwmueller (approved):
Lgtm

PR #168, opened 2023-05-02
Added save load functionality

sanjanag:
- Added save/load functionality and related tests
`Imagelab.save` saves `issues` and `issue_summary` as csv and clears them from imagelab instance on saving.
`Imagelab.load` warns on version incompatibility when loading
- `cleanvision.__version__` also works for python 3.7 now

### DIFF
--- a/tests/test_save_load.py
+++ b/tests/test_save_load.py
@@ -14,9 +14,20 @@ class TestImagelabSaveLoad:
         assert os.path.exists(save_folder / "issues.csv")
         assert os.path.exists(save_folder / "issue_summary.csv")
 
-    def test_force_save(self, generate_local_dataset, tmp_path):
+    @pytest.mark.parametrize(
+        "force",
+        [True, False],
+        ids=["overwrite", "do not overwrite"],
+    )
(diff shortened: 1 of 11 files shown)


tests/test_save_load.py
```diff
@@ -0,0 +1,66 @@
+import os
+
+import pytest
+
+import cleanvision
+from cleanvision.imagelab import Imagelab
+
+
```
jwmueller, line 8: add to one of the unit tests that actually runs imagelab:

imagelab.save(file)

imagelab2 = imagelab.load(file)

assert issues, issue_summary, info, and other attributes match between imagelab and imagelab2

jwmueller: afaict, none of the current tests actually checks the load recovers the saved object exactly.

jwmueller (changes requested):
LGTM but needs stronger unit tests that actually check for correct behavior

jwmueller (approved):
LGTM! Feel free to merge whenever you feel comfortable with it, as I think Elias may be too busy with datalab launch

Issue #82, opened 2023-02-01
New issue type: Detect abnormally big/small images


jwmueller:
Could also be based on width/height alone.

Threshold to determine `is_issue` (for overly large images) can be set at:  
size(image) > T * median(size(image) for image in dataset)
where T = 10 say. We don't want threshold to be percentage based, since that would always flag some images in any dataset.

The score could just be the percent of images that are bigger/smaller than this image.
Or could be size(image) / size(median_image) suitably renormalized to [0,1]

jwmueller:
Completed in https://github.com/cleanlab/cleanvision/pull/175

PR #175, opened 2023-05-11
Feat size: Added detection for images which are way bigger than average.

wirthual:
This adds functionality to detect images which are larger by a factor x (called ratio in the params) than the average image size.

Currently it is split up in width and height, but that's something which can be discussed. A possible extension would be to also detect very small images in comparison to the avg. image size.

Please let me know what you think.

This is work towards closing issue #82


wirthual:
I fixed typing and flake8 issues. Also increased the expected column count by 8 in the `run_test.py`. Additional columns are for now:
```['width','height', 'width_score_raw', 'is_width_issue', 'width_score','height_score_raw', 'is_height_issue','height_score']```

**Comment by sanjanag** [comment]
Hi @wirthual ! Great to see your initiative on picking up the issues and making relevant contributions.
I have a few questions regarding this PR
1. What dataset or set of images did you use to test this issue type? Could you post some results on this?
2. Follow up question: how did you decide the parameter `"ratio": 5.0`
3. Did you test with median instead of mean? Usually median is considered to be a better metric since mean is sensitive to outliers

Some suggestions and overview comments:
- This check could be included as part of image_property itself rather than a new issue manager. Let me know if you need any help in understanding the details of `image_property_issue_manager`
- The` imagelab.issues` and `issue_manager.issues dataframe` should only contains 2 columns related to an issue, `is_size_issue` `size_score`, rest of the useful information should go inside the `imagelab.info` and `issue_manager.info`
- Maybe try testing with area of the image, since it would encompass both height and width, height and width separately are anyway considered in the `odd_aspect_ratio` issue
- For getting both extremely large and small images in a dataset, you could use the this ratio for scoring `max(image_raw_score/image_median_score, image_median_score/image_raw_score)`

**Comment by wirthual** [comment]
Hi @sanjanag ,

To your questions:
1. I used the [Caltech-256 Object Category Dataset](https://authors.library.caltech.edu/7694). Those are the result with the default settings:
```
|    | issue_type       |   num_images |
|---:|:-----------------|-------------:|
|  0 | grayscale        |          422 |
|  1 | odd_aspect_ratio |          208 |
|  2 | blurry           |          204 |
|  3 | near_duplicates  |          167 |
|  4 | low_information  |          116 |
|  5 | width            |           91 |
|  6 | height           |           77 |
|  7 | dark             |           48 |
|  8 | exact_duplicates |           31 |
|  9 | light            |           17 | 
```
This means 91 images have a larger width than `5*mean_width` and 77 have a larger height than `5*mean_height`.
 
2.  The ratio of 5 I decided since I think that's a reasonable assumption to say if its 5 times larger than average it's too large/wide. In the issue, I think the number suggested was 10, but it seemed to be very high. 

3. I did not experiment with the median, but I like the idea and will have a look.

Regarding your comments:

I was unsure if this is an image property since it's not only depending on the image itself but rather on the full dataset. The mean/median of the sizes needs to be known beforehand in order to make an assumption.

Regarding area, this would mean we do not differentiate between width and height. If that's no valid use case, switching to an area is an excellent idea. But that's the main drawback I can think of.

**Comment by sanjanag** [comment]
Thank you for posting the results @wirthual ! I'd say it'd be more suited if we put this check also as part of `ImagePropertyIssueManager since the width and height are essentially image properties.

I think it's fine we don't differentiate between width and height because abnormal width and height of an image can be captured by `odd_aspect_ratio` and the rest can be covered by area.
It'd be great if you can get some results using the area property.

You could also use the example dataset for testing purposes and add a huge and tiny image to test your code. 
`wget -nc 'https://cleanlab-public.s3.amazonaws.com/CleanVision/image_files.zip'`

**Comment by wirthual** [comment]
Hi @sanjanag ,
thank you for your response. I will run my tests on the provided dataset and add a very large and a very small image.

If I understand the request correctly, I should include the behavior of the `size_issue_manager.py` into the `image_property_issue_manager.py` ?

I am not sure how to deal with the fact that I need to have the average image size of a dataset available before I can calculate the actual scores. So I would need to have the mean size available when I iterate over all the images.

**Comment by sanjanag** [comment]
> Hi @sanjanag , thank you for your response. I will run my tests on the provided dataset and add a very large and a very small image.
> 
> If I understand the request correctly, I should include the behavior of the `size_issue_manager.py` into the `image_property_issue_manager.py` ?

That's correct, @wirthual

**Comment by wirthual** [comment]
Changes:
 - refactored the code so the size check is now part of the image property checks
 - Size check is now based on the value of width * height
 - Size check now also detects very small images
 - I adapted the default threshold to be 10 as mentioned in the issue description
 - Added basic run tests to check the functionality
 
Here are the results of the code on the test set you provided with the default threshold of 10.0: 
```
Issues found in order of severity in the dataset
|    | issue_type       |   num_images |
|---:|:-----------------|-------------:|
|  0 | grayscale        |           20 |
|  1 | near_duplicates  |           20 |
|  2 | exact_duplicates |           19 |
|  3 | size             |           16 |
|  4 | dark             |           13 |
|  5 | blurry           |           10 |
|  6 | odd_aspect_ratio |            8 |
|  7 | light            |            5 |
|  8 | low_information  |            4 | 
```
With the output of the following flagged images (Because the score is lower than 0.1):
File | Size Score | Is Size Issue
-- | -- | --
/home/user/image_files/image_592.png | 0.0130061474368745 | True
/home/user/image_files/image_119.png | 0.015625 | True
/home/user/image_files/image_124.png | 0.015625 | True
/home/user/image_files/image_182.png | 0.015625 | True
/home/user/image_files/image_228.png | 0.015625 | True
/home/user/image_files/image_250.png | 0.015625 | True
/home/user/image_files/image_255.png | 0.015625 | True
/home/user/image_files/image_3.png | 0.015625 | True
/home/user/image_files/image_388.png | 0.015625 | True
/home/user/image_files/image_43.png | 0.015625 | True
/home/user/image_files/image_478.png | 0.015625 | True
/home/user/image_files/image_481.png | 0.015625 | True
/home/user/image_files/image_51.png | 0.015625 | True
/home/user/image_files/image_538.png | 0.015625 | True
/home/user/image_files/image_64.png | 0.015625 | True
/home/user/image_files/image_487.png | 0.0916392599055865 | True

This corresponds to the following images which indeed are very large or very small compared to the others.
image

File: tests/test_run.py [hunk 1]
```diff
@@ -206,5 +208,100 @@ def test_filepath_dataset_run(generate_local_dataset, images_per_class):
...
+    img = Image.fromarray(arr, mode="RGB")
+    img.save(Path(generate_local_dataset_once / "class_0" / "larger.png"))
+
+    files = os.listdir(generate_local_dataset_once / "class_0")
+    filepaths = [
+        os.path.join(generate_local_dataset_once / "class_0", f) for f in files
+    ]
+    imagelab = Imagelab(filepaths=filepaths)
+    imagelab.find_issues()
+    imagelab.report()
```
sanjanag, line 250: you could get rid of `imagelab.report()` in all the tests, as it's not being used anywhere

sanjanag:
Great work @wirthual ! The PR looks solid and I appreciate the unit tests you have added to check for specific cases.

We need to calibrate the threshold though, 10 was a suggestion it could be higher/lower. What we do usually is run the check on different datasets and see if the results look meaningful. The benchmark datasets we use are:
[Caltech-256](https://data.caltech.edu/records/nyy15-4j048)
[Food-101](https://data.vision.ee.ethz.ch/cvl/datasets_extra/food-101/)
[Caltech-UCSD-Birds](https://www.vision.caltech.edu/datasets/cub_200_2011/)
Could you please post results on above 3 datasets?

Feel free to choose any other datasets as well, that seem appropriate for this isssue check. Some resources for the same: [Hugginface](https://huggingface.co/datasets, [Kaggle](https://www.kaggle.com/datasets), [PapersWithCode](https://paperswithcode.com/datasets)

To save time, just run `imagelab.find_issues(issue_types: {"size": {}})`, and just run `imagelab.find_issues(issue_types: {"size": {"threshold": x}})` where say x=10 to check for different thresholds. You can try multiples of 5 for threshold, to keep the exercise simple and not too time consuming.

Once you have some results on that, it's good to go.

Also, mypy and flake8 are giving some errors, could you fix them? Let me know if you need any help in running these checks on your local system.

wirthual:
Here the results for [Caltech-256](https://data.caltech.edu/records/nyy15-4j048):

Threshold 5
|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | size         |         1830 | 

Threshold 10
|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | size         |          505 | 

Threshold 15
|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | size         |          337 | 

Threshold 20
|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | size         |          223 | 

Threshold 25
|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | size         |          136 | 


Top 4 examples with size issue in the dataset. (Same for all thresholds.)
image

sanjanag:
Can you post the results for Food-101 as well? @wirthual

sanjanag:
Hi @wirthual! Thank you for posting the results. The results from caltech dataset look interesting. We should not have a very low threshold like 2 or 5, as it would result in a lot of false positives.

I also think we should use `sqrt(area)` instead of the raw number, for a couple of reasons. First, you can imagine area can grow up pretty quickly and can potentially cause a memory issue. Second, it also makes more sense in terms of threshold, when you say a 10 times bigger image.

I would also suggest renaming the issue type from `size` to `odd_size` as that is more intuitive in what the issue is trying to capture.

To summarize:
- Rename the issue type
- use sqrt(area) instead of area
- start with a threshold of 10, utpo 25 is a good range

Could you please make the above changes and post the results for test dataset and caltech dataset again? Thank you.

wirthual:
Results for [Caltech-256](https://data.caltech.edu/records/nyy15-4j048):

Threshold 10

|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | odd_size     |            6 | 


Top 4 examples with odd_size issue in the dataset.
image

Threshold 15
|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | odd_size     |            1 | 

Found 1 example with odd_size issue in the dataset.
image

Threshold 20

|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | odd_size     |            1 | 


Found 1 example with odd_size issue in the dataset.
image
Threshold 25

|    | issue_type   |   num_images |
|---:|:-------------|-------------:|
|  0 | odd_size     |            1 | 

Found 1 example with odd_size issue in the dataset.
image

README.md
```diff
@@ -78,16 +78,17 @@ use the [cleanlab](https://github.com/cleanlab/cleanlab/) package.
...
+| 7 | Grayscale        | Images lacking color                                            | grayscale        | ![](https://raw.githubusercontent.com/cleanlab/assets/master/cleanvision/example_issue_images/grayscale.jpg)                                                                                                                                                        |
+| 8 | Odd Aspect Ratio | Images with an unusual aspect ratio (overly skinny/wide)        | odd_aspect_ratio | ![](https://raw.githubusercontent.com/cleanlab/assets/master/cleanvision/example_issue_images/odd_aspect_ratio.jpg)                                                                                                                                                 |
+| 9 | Size      | Images which are n times larger or smaller than the median size | size             | ![](https://raw.githubusercontent.com/cleanlab/assets/master/cleanvision/example_issue_images/odd_aspect_ratio.jpg) <img src="https://raw.githubusercontent.com/cleanlab/assets/master/cleanvision/example_issue_images/odd_aspect_ratio.jpg" width=20% height=20%> |
```
sanjanag, line 91: Change name of issue type here to `Odd Size`, also suggest a good example image to use for this issue type

src/cleanvision/issue_managers/image_property.py
```diff
@@ -324,6 +330,48 @@ def mark_issue(
         return is_issue
 
 
+class OddSizeProperty(ImageProperty):
+    name = "odd_size"
```
sanjanag, line 334: you can keep the property name `size`

src/cleanvision/issue_managers/image_property.py
```diff
@@ -324,6 +330,48 @@ def mark_issue(
         return is_issue
 
 
+class OddSizeProperty(ImageProperty):
```
sanjanag, line 333: Here too, keep the name SizeProperty

src/cleanvision/utils/constants.py
```diff
@@ -2,13 +2,15 @@
 
 IMAGE_PROPERTY: str = "image_property"
 DUPLICATE: str = "duplicate"
+SIZE: str = "size"
```
sanjanag, line 5: Looks like an unused constant

sanjanag:
@wirthual I've added my comments, can you update this branch with the base branch and make sure all tests pass. Thank you.

src/cleanvision/utils/constants.py
wirthual: Removed.

src/cleanvision/issue_managers/image_property.py
wirthual: Done.

src/cleanvision/issue_managers/image_property.py
wirthual: Done.

README.md
wirthual: Changed the name. What do you think would be a good example image?

sanjanag: I would suggest a very small image, since we can't show a very large image due to space constraints. You could resize one of the images in the test dataset to a very small size and add it here.

wirthual: Hi, I simply used one of the used images and rescaled it in the markdown.

Looks like this now:
image

I think the example images are hosted in another repo if I am not wrong.

sanjanag: Could you use a different image since this one is already used in the examples? Please raise a PR in this repo https://github.com/cleanlab/assets/tree/master/cleanvision/example_issue_images for example image

wirthual: Done:
https://github.com/cleanlab/assets/pull/3

sanjanag: Merged

wirthual: image

sanjanag (approved):

Issue #149, opened 2023-04-08
Support for large-scale datasets on the cloud


LemurPwned:
Hey,
first of all thanks for the great tool :) 

I think, it would be nice if the image reading capacity would be extended to non-local paths e.g. paths exposed via [`fsspec`-like interface types](https://filesystem-spec.readthedocs.io/en/latest/).

# Details
* Code/Notebooks running on the cloud could have easy access to the dataset which "looks" local without having to mount it explicitly.
For instance, in the function [`compute_scores`](https://github.com/cleanlab/cleanvision/blob/7f6b695f1371e91d163cb2349cc07a6c862da42a/src/cleanvision/issue_managers/image_property_issue_manager.py#L31):
```python
def compute_scores(
    fs: fsspec.spec.AbstractFileSystem, 
    path: str, to_compute: List[str], properties: Dict[str, ImageProperty]
) -> Dict[str, Any]:
    # image = Image.open(path) # this only opens local, not even a signed url
    with fs.open(path) as f:
        image = Image.open(f)
    results: Dict[str, Any] = {}
    results["path"] = path
    for issue_type in to_compute:
        results = {**results, **properties[issue_type].calculate(image)}
    return results
```
* Currently, for a large dataset, it needs to be either downloaded beforehand or directly mounted in the FS system.
* Could be super handy for large-scale unsupervised datasets.

jwmueller:
Great suggestion @LemurPwned !
Might you be interested in helping us implement this?

LemurPwned:
@jwmueller sure, I'll sketch something out :)

jwmueller:
Awesome! I am confident this would be super useful

jwmueller:
Related request from the [Cleanlab Community Slack](https://cleanlab-community.slack.com/archives/C03ENS0QQHJ/p1684028943984449?thread_ts=1683931381.099539&cid=C03ENS0QQHJ):    

Pass in a list of urls to images and run CleanVision

jwmueller:
Completed in https://github.com/cleanlab/cleanvision/pull/176

PR #176, opened 2023-05-11
Adding support for fsspec datasets

LemurPwned:
Hey,
so this is in response to #149, i.e. adding support for fsspec-like interfaces. List of supported fsspec implementations may be found here:
https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations 
Follow the exact fs resolution code here: https://github.com/fsspec/filesystem_spec/blob/4953f476944fdbae9e4e0c70930d77c7df2cdd85/fsspec/registry.py#L60 

I'm leaving it as draft for a second since I think it needs a bit of discussion. I tested the code with S3 bucket datasets and also locally. 

### Caveats 

Here is a list of caveats I have so far:
1. `fsspec.core.url_to_fs` returns a fs instance and this can also be a local instance. In this draft I've created a separate dataset class, but in principle the distinction between the local file system (represented by the `FolderDataset` class) and other filesystems need not exist, and it can be resolved by the protocol prefix, e.g. `s3://`, `gs://`. I didn't want to break existing stuff, so right now you can use the `fsspec` by providing `fsspec_dataset="s3://....` parameter with a given protocol.
2. Testing is done on local filesystem (default, local fs protocol). May be worth to include a self-hosted bucket with a small dataset, also useful for examples? 
3. When there is a huge number of images (~order of couple million or more), listing or globbing can take long time. I've found that using native libraries (boto3, google-cloud etc.) speeds this up significantly. That would only require reimplementing `get_filepaths` to use specific native libraries for listing & pattern matching (`glob` is not supported in most of them, only prefixes). But this is perhaps a very rare problem. 
4. Due to an [issue in fsspec 1019](https://github.com/fsspec/filesystem_spec/issues/1019), there's a reimplementation of `get_filepaths` function in the `FSDataset`. 

### Usage

If you have `aws`/`az`/`gs-utils` etc. configured then the fsspec interface should use this configuration automatically without having to pass the credentials explicitly (explicit secret passing is discouraged anyways). 
```python
imagelab = Imagelab(fsspec_dataset="s3://my-private-dataset/")
imagelab.find_issues()
```

Let me know what you think -- as a baseline should work. I'll probably add a simple example after the caveats are discussed.


sanjanag:
Hi @LemurPwned ! Thank you for the amazing work.
1. Regarding creating a separate argument in `Imagelab(fsspec_dataset)`, we can use the same `data_path` and `filepaths` argument for taking in fsspec paths and handle them internally according to the storage system. @jwmueller do you have any thoughts on this API change?
2. For testing, sure we can create a bucket that can be used to test this funtionality.
3. I think it is okay to use specific libraries for handling different storages, as long as the required dependencies are added. Just supporting the prefix for this case is fine IMO.

sanjanag:
Hi @LemurPwned !
After more thinking and deep diving into what fsspec does, here are my suggestions:
1. As mentioned earlier, we can use the same `data_path` and `filepaths` arguments for taking in fsspec paths and handle it internally in FolderDataset.
2. Regarding adding separate dependencies for handling different storages, I believe it will unnecessarily add a lot of dependencies to the repo and that will create too many dependencies to maintain.

Some other required changes:
1. The version of fsspec needs to be downgraded in dependencies that supports python 3.7 as well.
2. Could you specify extra depdencies required for s3 file system in `fsspec[s3]`

Could you please update the PR with above changes?
Thank you!

LemurPwned:
@sanjanag 
Thanks for the review!
Ad 1. Yes, this is not a problem -- if that internal change is acceptable, then I think it's the optimal implementation. I'll make the change.
Ad.2. I agree, but then I'm a bit confused about your later remark to add `fsspec[s3]` specifically. If the user uses s3 protocol path, there's a message to install s3fs, likewise for other protocols from known implementations. 
Ad. `fsspec` version -- I'll find the version that's also available for 3.7.

LemurPwned:
@sanjanag yes! It totally makes sense -- I was just wondering why s3 in particular, i.e. there's `fsspec[gcs]` or `fsspec[adl]` too.

sanjanag:
I mentioned s3 just from a testing perspective, first let's get the s3 done and then others will follow. In the same PR or a separate PR, you could then add support for google cloud and azure.

LemurPwned:
@sanjanag just to be clear -- this PR and changes made here already added support for all the known implementations I've listed (https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations). I just happened to test on `s3`.

I'll make the changes and add s3fs to the requirements, but if it's just the testing you worry about, I can test on some other solutions too.

sanjanag:
Hi @LemurPwned ! I understand the PR changes already support different implementations. Could you add dependencies for s3, azure and google cloud and test them as well? Thank you.

sanjanag:
Hi @LemurPwned! Are there any updates on this PR? Let me know if I can help in pushing this change forward.

LemurPwned:
Hi @sanjanag -- sorry! I've been somewhat busy past two weeks but found a spare minute. 

Here's the current status:
- I removed `FolderDataset` and replaced it with the instance of `FSDataset` which operates both locally and on individual protocols.
- For Azure support I needed to expose additional `storage_opts`, a dict where the account name and the account key are specified.
- Added the extra requirements for the azure and s3 protocols. Separated the requirements for python3.7 and above. Tested the latter on `python:3.7-buster` so test should pass now. 
- Unfortunately, I don't have GC student billing now but I remember that s3fs and gcsfs are identical, so if s3fs worked, so will gcsfs (unlike adfs which requires extra params, as I mentioned above). It means that it's supported out of the box as-is, but I simply haven't tested it yet.

Example for testing Azure dataset:
```python
ACCOUNT_KEY = os.environ.get("AZURE_STORAGE_ACCOUNT_KEY")
storage_opts = {'account_name': 'cleanstorage', 'account_key': ACCOUNT_KEY}
imagelab = Imagelab(data_path=folder, storage_opts=storage_opts)
imagelab.find_issues()
```
and s3/gs
```python
folder = "s3://my-private-dataset/" # or "gs://my-private-dataset/" 
imagelab = Imagelab(data_path=folder)
imagelab.find_issues()
```

Let me know if those changes are in line of what you expected.

pyproject.toml
```diff
@@ -36,13 +36,22 @@ dependencies = [
     "tabulate>=0.8.3", # pandas optional dependency for .to_markdown()
     "imagehash>=4.2.0",
     "tqdm>=4.53.0",
-    "matplotlib>=3.4"
+    "matplotlib>=3.4",
+    "fsspec==2023.1.0; python_version<'3.8'", # latest compatible with Python 3.7
```
sanjanag, line 40: Why the latest compatible and not the minimum compatible version here?

sanjanag:
Hi @LemurPwned ! Thanks for pushing out the updated changes.

Could you add a jupyter notebook on how to run CleanVision for a dataset residing in s3, azure and gs [here](https://github.com/cleanlab/cleanvision/tree/main/docs/source/tutorials) similar to how you have done in the code snippets above. This would be super helpful in reviewing the PR as well as users who are going to interact with the code. Include details like whether you need to authenticate or it will pick up default credentials.

Add some unit tests ensuring cleanvision is being run on files in cloud storage and producing some results.

pyproject.toml
LemurPwned: Could you explain the difference? 
What I meant is that later versions do not support python 3.7. If the user's python version is above 3.7, it's installing the minimum compatible version.

sanjanag:
Sure @LemurPwned we do not expect that you use your personal accounts. Sorry, did not make that explicit in the last message. For s3 I can setup a public bucket, since we have the access. It would be great if you could find some relevant public datasets on google cloud and azure. I will also look for them.

LemurPwned:
@sanjanag I've found a small dataset `s3://amazon-berkeley-objects/images/small` from the AWS Open Datasets.
I'm using a selected portion of it for the testing and there's a notebook example now that uses this datasets as an example.
I've left the implicit authentication with closed dataset.

LemurPwned:
Hi @sanjanag, I've marked it read y for review -- happy to make any remaining fixes.
I also tested it on the gcsfs too -- it works :) 

There is only one issue -- for large datasets on cloud this can be quite slow. I have some ideas how to fix it, but it'll require more in-depth digging, perhaps in a second PR.

sanjanag:
@LemurPwned The major part of this PR is done, thank you for your contribution. If it's okay with you, I can take this PR from here and to make it merge ready.

LemurPwned:
@sanjanag thank you -- yes, no worries :) Let me know when there's anything to do regarding fsspec.

sanjanag:
Hi @LemurPwned ! Could you help with [this issue](https://github.com/cleanlab/cleanvision/pull/176#issuecomment-1621998691) when you get time? Thanks!

LemurPwned:
Hi @sanjanag, sorry! I completely missed your mention. That was my bug -- I was running with an aws access configured locally so it always defaulted to that config. I now added `anon` access which should not require any configuration for open s3 buckets. 


Let me know if that fixes the issue for you. Sorry again!

src/cleanvision/dataset/fsspec_dataset.py
```diff
@@ -18,11 +18,12 @@ def __init__(
         self,
         data_folder: Optional[str] = None,
         filepaths: Optional[List[str]] = None,
-        storage_opts: Dict[str, str] = {},
+        storage_opts: Optional[Dict[str, str]] = {},
     ) -> None:
         super().__init__()
         self.storage_opts = storage_opts
-        ignore_missing = self.storage_opts.pop("ignore_missing_keys", False)
+        if self.storage_opts:
```
LemurPwned: I think this if is not required with `False` being a default for the `pop` function?

jwmueller (approved):
LGTM. Thanks @LemurPwned!

PR #177, opened 2023-05-12
Minor updates

sanjanag:
- Updated progress bar
New:
Screenshot 2023-05-12 at 12 16 50 PM
Old: this one used to split sometimes into multiple lines in jupyter notebooks
Screenshot 2023-05-12 at 12 17 28 PM
- Add try/catch in case image cannot be read from a huggingface dataset, encountered this in a couple of datasets datasets
This skips issue checking over images which could not be loaded and assigns Nan scores for all image property issues and 0 for duplicate issues, overall `is_issue` columns are set to `False` for these corrupted images.
Screenshot 2023-05-15 at 1 57 48 PM

### DIFF
--- a/src/cleanvision/dataset/hf_dataset.py
+++ b/src/cleanvision/dataset/hf_dataset.py
@@ -33,6 +33,7 @@ class HFDataset(Dataset):
             return image
         except Exception as e:
             print(f"Could not load image at index: {item}", e)
+            return None
 
     def _set_index(self) -> None:
         self.index = [i for i in range(len(self._data))]
--- a/src/cleanvision/dataset/hf_dataset.py
+++ b/src/cleanvision/dataset/hf_dataset.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Union
+from typing import TYPE_CHECKING, Union, Optional
 
 from PIL import Image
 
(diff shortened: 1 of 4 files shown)


src/cleanvision/dataset/hf_dataset.py
```diff
@@ -27,8 +27,13 @@ def __init__(self, hf_dataset: "datasets.Dataset", image_key: str):
...
 
-    def __getitem__(self, item: Union[int, str]) -> Image.Image:
-        return self._data[item][self._image_key]
+    def __getitem__(self, item: Union[int, str]) -> Optional[Image.Image]:
+        try:
+            image = self._data[item][self._image_key]
+            return image
+        except Exception as e:
+            print(f"Could not load image at index: {item}", e)
+            return None
```
jwmueller, line 36: would be good if you can update the PR description to clarify at the Imagelab level what the user experience is like when there is an image like this in the dataset.

Does the entire call to find_issues fail or does it just skip over this image?
The latter behavior would be desirable in future if it's not implemented, but need to think about how to score such an image / other details

sanjanag: Added details about the change in description

jwmueller (approved):
Looks great! Thanks for updating the description

PR #181, opened 2023-05-20
Truncate long titles in plots

### LINKED ISSUE #122: Handle long filenames in visualization
Screenshot 2023-03-14 at 3 48 36 PM

manulpatel:
Fixes #122 

1. We determine direction of truncation using longest common substring on first 2 titles of dataset. Length of two longest substrings are compared by truncating both titles, one from front and other from back. The greater the length of substring, that direction is applied wherver required futher.

2. Using a for loop, we check, each image's `title width >= cell width`, by bringing both in inches units.

3. If Step-2 condition is true then, by following the truncation direction from step1 we slice the title to `(cell_size[0] /0.09722) - 4` characters and further add "..." at  front/end.

**Before:**

**After:**


jwmueller:
Thanks for the PR on this!  

Could you please include a screenshot of the before/after for a dataset with some long filenames, but not every filename being long?

src/cleanvision/utils/viz_manager.py
```diff
@@ -45,6 +45,12 @@ def plot_image_grid(
     fig, axes = plt.subplots(
         nrows, ncols, figsize=(cell_size[0] * ncols, cell_size[1] * nrows)
     )
+
+    for i in range(len(images)):
+        title_width = len(titles[i]) * 0.09722
```
jwmueller, line 50: please comment code to describe this constant and how it was chosen

jwmueller: it should also be defined as a variable like `CHARACTER_SIZE`, and then you should reference this variable below instead of repeating the constant. This makes it easier to change the value of the constant in the future

manulpatel: Sure! I felt this would be helpful for future developers.

src/cleanvision/utils/viz_manager.py
```diff
@@ -45,6 +45,15 @@ def plot_image_grid(
...
     )
+
+    for i in range(len(images)):
+        """Converts font size of 7 into inches"""
+        CHARACTER_SIZE_INCHES = 7 * (1 / 72)
+
+        title_width = len(titles[i]) * CHARACTER_SIZE_INCHES
+        if title_width >= cell_size[0]:
+            titles[i] = titles[i][: math.ceil(cell_size[0] / CHARACTER_SIZE_INCHES) - 1]
+
```
jwmueller, line 56: thanks for the update!  One further improvement IMO will be the following:
After the `if` clause, instead truncate the `titles[i]` a little more than necessary, so that you can append the suffix "..." to the end of the string to indicate it was truncated. 

Please update your main screenshot in PR description with some images where the title is truncated AND some images where the title is not truncated (in the same dataset). With the "..." suffix, it should become more obvious which image paths were truncated vs not.

jwmueller: The more advanced version of this that I'd encourage you to consider for this or a future PR, is to do the truncation from the beginning or from the end, whichever makes the truncated strings more unique.  

Currently with the truncation always from the end, you can see there is no way for user to distinguish the file names anymore in your dataset pictured here. When you consider truncating from beginning as well, you can simply add the "..." as a prefix instead of as a suffix.

So for datasets where only the end of the filenames differ, we'd prefer them to be truncated at their beginning.

manulpatel: Yes I would be very happy to implement the advanced version. I have started a thread in slack to discuss further on this.

manulpatel: Done! Updated in the new commits.

jwmueller: thanks! Unit tests are failing though, can you fix and add new unit test for your new functionality?

manulpatel: Updated. Handled the case of 1 image in dataset.

jwmueller (approved):
Thanks for the nice contribution!

PR #182, opened 2023-05-22
Changed import for Imagelab

sanjanag:
Now one can import like this:
`from cleanvision import Imagelab`

### DIFF
--- a/src/cleanvision/__init__.py
+++ b/src/cleanvision/__init__.py
@@ -1,4 +1,5 @@
 import sys
+from cleanvision.imagelab import Imagelab as _Imagelab
 
 PYTHON_VERSION_INFO = sys.version_info
 
@@ -18,3 +19,5 @@ try:
     __version__ = get_version()
 except Exception:
     pass
+
+Imagelab = _Imagelab
--- a/tests/test_save_load.py
+++ b/tests/test_save_load.py
@@ -5,7 +5,9 @@ import pytest
 from pandas.testing import assert_frame_equal, assert_series_equal
 
 import cleanvision
-from cleanvision import Imagelab
+
+# To ensure backwards compatibility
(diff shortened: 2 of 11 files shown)


src/cleanvision/__init__.py
```diff
@@ -18,3 +19,5 @@ def get_version() -> str:
     __version__ = get_version()
 except Exception:
     pass
+
+Imagelab = _Imagelab
```
jwmueller, line 23: just curious why this is necessary instead of just `from cleanvision.imagelab import Imagelab` ?

tests/test_save_load.py
```diff
@@ -5,7 +5,7 @@
 from pandas.testing import assert_frame_equal, assert_series_equal
 
 import cleanvision
-from cleanvision.imagelab import Imagelab
+from cleanvision import Imagelab
```
jwmueller, line 8:
```suggestion
from cleanvision.imagelab import Imagelab
```

jwmueller: how about we keep the old way of importing for here, to ensure backwards compatibility in the future?

jwmueller (approved):
LGTM, don't forget to update our blogpost and cleanvision-examples notebook repos too.
Don't merge this PR until right before you are ready to release the next cleanvision version to avoid confusing people.

sanjanag:
There's some error in building docs, I'm fixing that

tests/test_save_load.py
sanjanag: We can do that

src/cleanvision/__init__.py
sanjanag: It works, but flake8 throws an error for unused import. We should keep it in case someone accidentally removes the import statement through editor as it is unused.

PR #183, opened 2023-05-23
Added score for duplicate images

sanjanag:
Added score for exact_duplicates and near_duplicates issue type.

`score = 1 / num_images_in_duplicated_set`

### DIFF
--- a/src/cleanvision/issue_managers/duplicate_issue_manager.py
+++ b/src/cleanvision/issue_managers/duplicate_issue_manager.py
@@ -184,7 +184,9 @@ class DuplicateIssueManager(IssueManager):
         for issue_type in self.issue_types:
             score_col = get_score_colname(issue_type)
             for s in self.info[issue_type][SETS]:
-                score = 1.0 / len(s)  # will never be 0 because all images in this set are duplicated
+                score = 1.0 / len(
+                    s
+                )  # will never be 0 because all images in this set are duplicated
                 score_df.loc[s, score_col] = score
 
             self.issues = self.issues.join(score_df[[score_col]])
--- a/src/cleanvision/issue_managers/duplicate_issue_manager.py
+++ b/src/cleanvision/issue_managers/duplicate_issue_manager.py
@@ -184,7 +184,7 @@ class DuplicateIssueManager(IssueManager):
         for issue_type in self.issue_types:
             score_col = get_score_colname(issue_type)
             for s in self.info[issue_type][SETS]:
-                score = 1.0 / len(s)
+                score = 1.0 / len(s)  # will never be 0 because all images in this set are duplicated
                 score_df.loc[s, score_col] = score
 
             self.issues = self.issues.join(score_df[[score_col]])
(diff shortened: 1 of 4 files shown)


src/cleanvision/issue_managers/duplicate_issue_manager.py
```diff
@@ -173,15 +173,24 @@ def _update_summary(self) -> None:
...
+            score_col = get_score_colname(issue_type)
             for s in self.info[issue_type][SETS]:
-                duplicated_images.extend(s)
-            self.issues[
-                get_is_issue_colname(issue_type)
-            ] = self.issues.index.to_series().apply(
-                lambda x: True if x in duplicated_images else False
-            )
+                score = 1.0 / len(s)
+                score_df.loc[s, score_col] = score
```
jwmueller, line 188: can you cover this line in unit test, if easy?

tests/test_run.py
```diff
@@ -171,7 +171,7 @@ def test_hf_dataset_run(generate_local_dataset, n_classes, images_per_class):
     imagelab = Imagelab(hf_dataset=hf_dataset, image_key="image")
     imagelab.find_issues()
     imagelab.report()
-    assert len(imagelab.issues.columns) == 14
+    assert len(imagelab.issues.columns) == 16
     assert len(imagelab.issues) == n_classes * images_per_class
```
jwmueller, line 175: what is the most duplicated image in our testing dataset?  couldn't we easily add a test to verify this image has the lowest score (tied w other images in its duplicate-set) ?  The current tests don't seem to test any of the logic at all

src/cleanvision/issue_managers/duplicate_issue_manager.py
```diff
@@ -173,15 +173,24 @@ def _update_summary(self) -> None:
...
-            duplicated_images = []
+            score_col = get_score_colname(issue_type)
             for s in self.info[issue_type][SETS]:
-                duplicated_images.extend(s)
-            self.issues[
-                get_is_issue_colname(issue_type)
-            ] = self.issues.index.to_series().apply(
-                lambda x: True if x in duplicated_images else False
-            )
+                score = 1.0 / len(s)
```
jwmueller, line 187: what happens if `len(s) = 0`? Is this guaranteed to never happen? Would add a comment stating this assumption, since it is a strong one

sanjanag: It's never going to be 0, since all the sets in this list are duplicated sets.
It looks like this
`[[im_1, im2], [im_3, im_4,  im_5]]`

jwmueller, line 187:
```suggestion
                score = 1.0 / len(s)  # will never be 0 because all images in this set are duplicated
```

src/cleanvision/issue_managers/duplicate_issue_manager.py
sanjanag: Covered in unit test

tests/test_run.py
sanjanag: Covered in unit test

elisno (approved):
LGTM!

Cool that you're sorting the duplicate issues by the duplicate counts when visualizing.

PR #185, opened 2023-05-23
Blurry

sanjanag:
- Updated logic for blurry issue type
   Previous logic avoided FP like objects against black backgrounds by a low blurriness threshold. This also resulted in more FN. The current logic better handles these kinds of images, hence we can include decrease FN by increasing the threshold.
- Added an optional argument `indices` in `Imagelab.visualize()`, found this useful when evaluating huggingface and torchvision datasets.
Screenshot 2023-05-24 at 11 13 25 AM

### DIFF
--- a/src/cleanvision/issue_managers/image_property_issue_manager.py
+++ b/src/cleanvision/issue_managers/image_property_issue_manager.py
@@ -23,7 +23,6 @@ from cleanvision.utils.constants import (
 from cleanvision.utils.utils import (
     get_is_issue_colname,
     update_df,
-    get_score_colname,
 )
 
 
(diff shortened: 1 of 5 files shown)


src/cleanvision/issue_managers/image_property_issue_manager.py
```diff
@@ -66,7 +65,11 @@ def get_default_params(self) -> Dict[str, Any]:
                 "threshold": 0.3,
                 "normalizing_factor": 0.1,
             },
-            IssueType.BLURRY.value: {"threshold": 0.17, "normalizing_factor": 0.01},
+            IssueType.BLURRY.value: {
+                "threshold": 0.29,
+                "normalizing_factor": 0.01,
+                "color_threshold": 0.18,
```
elisno, line 71: No need to do it in code, just here in the discussion:

Why is the threshold bumped, but the color_threshold is similar to the old threshold?

elisno (approved):

src/cleanvision/issue_managers/image_property_issue_manager.py
sanjanag: Updated description

PR #201, opened 2023-06-21
Testing for windows

sanjanag:


ulya-tkch (approved):
If `3.7` works on windows we should stick to that instead of `3.7.9`. Otherwise LGTM! Reminder to change it back once https://github.com/actions/setup-python/issues/682 is addressed

PR #204, opened 2023-07-03
Updated development guide with formatting and styling commands

sanjanag:


ulya-tkch (approved):
LGTM! Commands same as in .ci file and flake8 is in requirements-dev.txt

PR #217, opened 2023-08-29
Fixed a time consuming line in odd size image property

sanjanag:


src/cleanvision/issue_managers/image_property.py
```diff
@@ -356,13 +356,10 @@ def get_scores(
...
         assert raw_scores is not None
+        median_score = raw_scores[self.score_columns[0]].median()
         scores = pd.DataFrame(index=raw_scores.index)
         scores[get_score_colname(issue_type)] = raw_scores[self.score_columns[0]].apply(
-            lambda x: 1.0
-            / max(
-                x / raw_scores[self.score_columns[0]].median(),
-                raw_scores[self.score_columns[0]].median() / x,
-            )
+            lambda x: 1.0 / max(x / median_score, median_score / x)
```
jwmueller, line 362: kind of odd to do the redundant division instead of just min( x / median, median / x) which to my understanding appears the same, but maybe I'm missing something

elisno:
I've conducted a thorough benchmark on various strategies, including the ones in the original code, your PR, and a few other optimized methods.

Here are the key takeaways:

- **Benchmarking and Equality**: I started by verifying that the output of all strategies was consistent for a smaller dataset. This ensured that any performance improvements didn't come at the expense of accuracy.

- **Performance**: The strategies were then benchmarked across various dataset sizes to see how they scale.

From the benchmarks, I observed that the "Optimized Min Strategy" demonstrated consistent and superior performance, especially as the dataset size grew.

Given the results, I'd suggest we move forward with that strategy. It offers a further boost in performance with better scaling.

675eaa0f-7b1f-4196-9665-529bf6751ad5

The Numba strategies are just a reference for further potential optimizations, but I don't suggest we use those now.

<details> <summary> Click here to see the code I used for the benchmarks </summary>

```python
import pandas as pd
import numpy as np
import timeit
import matplotlib.pyplot as plt
from typing import Callable
from numba import jit

# ----------- DATA GENERATION -----------
def create_random_data(N: int) -> pd.DataFrame:
    return pd.DataFrame({'size': np.random.rand(N)})


# ----------- BENCHMARK FUNCTION -----------
def benchmark(strategy: Callable, data: pd.DataFrame) -> float:
    wrapper = lambda: strategy(data)
    return timeit.timeit(wrapper, number=20) / 20


# ----------- EQUALITY CHECKS -----------
def check_equality(results: dict, tol=1e-8):
    strategies = list(results.keys())
    for i in range(len(strategies)):
        for j in range(i + 1, len(strategies)):
            assert np.allclose(results[strategies[i]], results[strategies[j]], atol=tol), \
                f"Results of {strategies[i]} and {strategies[j]} are not approximately the same!"


# ----------- STRATEGIES -----------
def original_strategy(data: pd.DataFrame) -> pd.Series:
    return data['size'].apply(lambda x: 1.0 / max(x / data['size'].median(), data['size'].median() / x))


def pr_strategy(data: pd.DataFrame) -> pd.Series:
    median_score = data['size'].median()
    return data['size'].apply(lambda x: 1.0 / max(x / median_score, median_score / x))


def optimized_strategy(data: pd.DataFrame) -> pd.Series:
    size_scores = data['size']
    median_score = size_scores.median()
    scores_ratio = size_scores / median_score
    
    # Computing the values of the two divisions
    division_1 = scores_ratio
    division_2 = 1.0 / scores_ratio

    # Using np.maximum to get the element-wise max values
    max_values = np.maximum(division_1, division_2)

    # Finally, returning the reciprocal of the max values
    return 1.0 / max_values

def optimized_min_strategy(data: pd.DataFrame) -> pd.Series:
    size_scores = data['size']
    median_score = size_scores.median()
    scores_ratio = size_scores / median_score
    
    # Computing the values of the two divisions
    division_1 = scores_ratio
    division_2 = 1.0 / scores_ratio

    # Using np.minimum to get the element-wise min values
    min_values = np.minimum(division_1, division_2)

    # No need to compute the reciprocal
    return min_values

@jit(nopython=True)
def compute_reciprocal(values, median_score):
    n = len(values)
    results = np.empty(n)
    
    for i in range(n):
        value = values[i]
        division_1 = value / median_score
        division_2 = median_score / value
        max_value = max(division_1, division_2)
        results[i] = 1.0 / max_value
        
    return results

def numba_strategy(data: pd.DataFrame) -> pd.Series:
    values = data['size'].values
    median_score = np.median(values)
    return pd.Series(compute_reciprocal(values, median_score))

@jit(nopython=True)
def compute_reciprocal_min(values, median_score):
    n = len(values)
    results = np.empty(n)
    
    for i in range(n):
        value = values[i]
        division_1 = value / median_score
        division_2 = median_score / value
        min_value = min(division_1, division_2)
        results[i] = min_value
        
    return results

def numba_min_strategy(data: pd.DataFrame) -> pd.Series:
    values = data['size'].values
    median_score = np.median(values)
    return pd.Series(compute_reciprocal_min(values, median_score))


# ----------- BENCHMARK ALL STRATEGIES -----------
strategy_config = {
    "Original Strategy": {
        "func": original_strategy,
        "threshold": 1e3
    },
    "PR Strategy": {
        "func": pr_strategy,
        "threshold": 1e5
    },
    "Optimized Strategy": {
        "func": optimized_strategy,
    },
    "Optimized Min Strategy": {
        "func": optimized_min_strategy
    },
    "Numba Strategy": {
        "func": numba_strategy,
    },
    "Numba Min Strategy": {
        "func": numba_min_strategy
    },
}

# Run and benchmark each strategy for a single dataset (for jit)
df = create_random_data(10)
results = {name: config["func"](df) for name, config in strategy_config.items()}

# Check equality for this dataset
check_equality(results)

# Benchmark for different dataset sizes
Ns = np.logspace(1, 7, 30).astype(int)
results_times = {strategy: [] for strategy in strategy_config}

for N in Ns:
    df = create_random_data(N)
    for name, config in strategy_config.items():
        threshold = config.get("threshold", None)
        
        if threshold is not None and N > threshold:
            results_times[name].append(None)
        else:
            results_times[name].append(benchmark(config["func"], df))


# ----------- VISUALIZATION -----------
plt.figure(figsize=(10, 7))
for name, times in results_times.items():
    valid_Ns = [N for N, t in zip(Ns, times) if t is not None]
    valid_times = [t for t in times if t is not None]
    plt.plot(valid_Ns, valid_times, label=name, marker='o')

plt.xscale('log')
plt.yscale('log')
plt.xlabel('Number of Samples (N)')
plt.ylabel('Execution Time (seconds)')
plt.title('Benchmarking Different Strategies')
plt.legend()
plt.grid(True, which="both", ls="--", c='0.65')
plt.show()

```

</details>

src/cleanvision/issue_managers/image_property.py
elisno: See [benchmarks](https://github.com/cleanlab/cleanvision/pull/217#issuecomment-1698037300).

elisno (approved):
Added some compute optimizations.

LGTM! Feel free to merge when CI goes green.

Issue #215, opened 2023-07-28
Extend CleanVision to run on video data


jwmueller:
Some discussion of this here:  https://github.com/cleanlab/cleanvision/discussions/214

Making CleanVision runnable on video would be super useful to many folks!

Should be fairly easy:

- just extract every k-th frame from the video,
- run cleanvision on those,
- aggregate the results across frames per video

LemurPwned:
I'd be super interested in this -- might come in handy when sampling for annotations etc. Potentially doing some on-the-fly deduplication using the perceptual hash to avoid sampling visually identical scenes, before passing in to do further analysis, could improve efficiency of the process.

jwmueller:
@LemurPwned This would be an awesome contribution!  Nobody is working on this actively AFAIK. 

@lafemme https://github.com/lafemmephile 
was interested in trying CleanVision for video and could test out your PR code too!

LemurPwned:
@jwmueller Great, I started playing around with the idea and for simplicity sake I created a separate repo for the sampling part: https://github.com/LemurPwned/video-sampler.
It seems that depending on the settings a significant reduction can take place. It's also pretty damn quick!

 I'll try to integrate the sampling with the clean vision api in a PR soon.

jwmueller:
awesome super excited about this!!

lafemmephile:
> I'd be super interested in this -- might come in handy when sampling for annotations etc. Potentially doing some on-the-fly deduplication using the perceptual hash to avoid sampling visually identical scenes, before passing in to do further analysis, could improve efficiency of the process.

I like the *deduplication* feature ... I saw the code you wrote ... looks very interesting. Seems like you've done some good work there, I want to study it more in depth.

LemurPwned:
@lafemmephile thanks, just to be clear, you want to pick up from here?

lafemmephile:
> @lafemmephile thanks, just to be clear, you want to pick up from here?

@LemurPwned  Yes. I'm going to take a stab at assembling this `Videolab` class/module, which is much easier than the work you've already done (again thank you so much for developing the video sampling solution).

With the work you've already done (i.e. the *frame sampling*), it seems we are tantalizingly close to a *minimum viable product* (i.e. something approximating `cleanvision.imagelab.Imagelab` but for *video data* aka `cleanvision.videolab.Videolab`).

The approach I am roughly considering:
[...]

PR #231, opened 2023-10-17
Videolab: Minimum Viable Class

lafemmephile:
# Abstract
The purpose of issue #215 was to implement a *minimal* extension of the `Imagelab` class. I believe that at this point, per the criteria listed [here](https://github.com/cleanlab/cleanvision/issues/215#issuecomment-1752008997), we are now very close to achieving the aforementioned *extension*. While I fully expect there will be *specific* issues that the `cleanvision` maintainers will want to address, the extension is mature enough to begin the review process.

# Implementation Strategy
In an effort to keep to the original goal of the extension laid out by @jwmueller, I have not attempted any additional features (including any *default arguments*, or error checking). I have only implemented exactly that which is necessary to fulfill the *three tenets*:

**1**. just extract every k-th frame from the video,
**2**. run cleanvision on all those images,
**3**. aggregate the results across the frames extracted per video, into scores per video.

Tenet **1** was implemented in quite a simple way, with the `FrameSampler` class (which is a simplified version of the [VideoSampler](https://github.com/LemurPwned/video-sampler) written by @LemurPwned). Again you will notice I limited any *excess features* except where necessary, in order to keep to the *minimal* design and limited scope of issue #215.

Tenet **2** and **3** were achieved with the `Videolab` class. Again, where possible I have limited any excess features or additional design choices, preferring to defer to the `cleanvision` maintainers for specifics.

# Conclusion
Thank you to @smttsp @LemurPwned @sanjanag and @jwmueller for their contributions and perspective on *extending Cleanvision to run on video data*.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ classifiers = [
 ]
 dependencies = [
     "Pillow>=9.3",
-    "av>=10.0.0",
     "numpy>=1.20.0",
     "pandas>=1.1.5",
     "tabulate>=0.8.3", # pandas optional dependency for .to_markdown()
@@ -47,11 +46,11 @@ pytorch = ["torchvision>=0.12.0"]
 azure = ["adlfs>=2022.2.0"] # latest compatible with Python 3.7
 gcs = ["gcsfs>=2022.1.0"] # latest compatible with Python 3.7
 s3 = ["s3fs>=2023.1.0"] # latest compatible with Python 3.7
+video = ["av>=10.0.0"]
 
-all = ["cleanvision[huggingface,pytorch,azure,gcs,s3]"]
+all = ["cleanvision[huggingface,pytorch,azure,gcs,s3,video]"]
 
 [project.urls]
 "Source" = "https://github.com/cleanlab/cleanvision"
 "Bug Tracker" = "https://github.com/cleanlab/cleanvision/issues"
 "Documentation" = "https://cleanvision.readthedocs.io/"
-
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ classifiers = [
 ]
 dependencies = [
     "Pillow>=9.3",
+    "av>=10.0.0",
     "numpy>=1.20.0",
     "pandas>=1.1.5",
     "tabulate>=0.8.3", # pandas optional dependency for .to_markdown()
--- a/src/cleanvision/videolab.py
+++ b/src/cleanvision/videolab.py
@@ -1,15 +1,64 @@
 """Videolab is an extension of Imagelab for finding issues in a video dataset."""
 from importlib import import_module
 from pathlib import Path
-from typing import Any, Dict, Generator, List, Optional
+from typing import Any, Dict, Generator, Iterator, List, Optional, Union
 
 import pandas as pd
 from PIL.Image import Image
+from tqdm.auto import tqdm
 
+from cleanvision.dataset.base_dataset import Dataset
 from cleanvision.imagelab import Imagelab
 from cleanvision.utils.utils import get_is_issue_colname
 
 VIDEO_FILE_EXTENSIONS = ["*.mp4", "*.avi", "*.mkv", "*.mov", "*.webm"]
+
+
+class VideoDataset(Dataset):
+    """Wrapper class to handle video datasets."""
+
+    def __init__(
+        self,
+        data_folder: Optional[str] = None,
+        filepaths: Optional[List[str]] = None,
+    ) -> None:
+        """Determine video dataset source and populate index."""
+        # check if data folder is given
+        if data_folder:
+            # get filepaths from video dataset directory
+            self._filepaths = [
+                str(path) for path in self.__get_filepaths(Path(data_folder))
+            ]
+
+        else:
+            # store user supplied video file paths
+            assert filepaths is not None
+            self._filepaths = filepaths
+
+        # create index
+        self._set_index()
+
+    def __len__(self) -> int:
+        """Get video dataset file count."""
+        return len(self.index)
+
+    def __iter__(self) -> Iterator[Union[int, str]]:
+        """Defining the iteration behavior."""
+        return iter(self.index)
+
+    def _set_index(self) -> None:
+        """Create internal storage for filepaths."""
+        self.index = [path for path in self._filepaths]
+
+    def __get_filepaths(self, dataset_path: Path) -> Generator[Path, None, None]:
+        """Scan file system for video files and grab their file paths."""
+        # notify user
+        print(f"Reading videos from {dataset_path}")
+
+        # iterate over video file extensions
+        for ext in VIDEO_FILE_EXTENSIONS:
+            # loop through video paths matching ext
+            yield from dataset_path.glob(f"**/{ext}")
 
 
 class FrameSampler:
@@ -29,37 +78,42 @@ class FrameSampler:
                 "Please install it via `pip install av` and then try again."
             ) from error
 
-    def _create_frame_sample_sub_dir(self, video_file: Path, output_dir: Path) -> Path:
+    def _create_frame_sample_sub_dir(self, output_dir: Path, idx: int) -> Path:
         """Create a unique sub direcotry for storing frame samples from a video file."""
-        # create new sub directory from video file name
-        sub_dir = output_dir / video_file.name
+        # create new sub directory from video_dataset index
+        sub_dir = output_dir / str(idx)
         sub_dir.mkdir(parents=True)
 
         # return path to new sub dir
         return sub_dir
 
-    def sample(self, video_file: Path, output_dir: Path) -> None:
+    def sample(self, video_dataset: VideoDataset, output_dir: Path) -> None:
         """Loop through frames and store every k-th frame."""
-        # create frame samples sub directory
-        sample_sub_dir = self._create_frame_sample_sub_dir(video_file, output_dir)
+        # notify of sampling
+        print(f"Sampling frames at every {self.k} frames ...")
 
-        # open video file for streaming
-        with self.av.open(str(video_file)) as container:
-            # get video stream
-            stream = container.streams.video[0]
+        # iterate over video files in video data directory
+        for idx, video_file in enumerate(tqdm(video_dataset)):
+            # create frame samples sub directory
+            sample_sub_dir = self._create_frame_sample_sub_dir(output_dir, idx)
 
-            # iterate frames
-            for frame_indx, frame in enumerate(container.decode(stream)):
-                # check for k-th frame
-                if not frame_indx % self.k:
-                    # get PIL image
-                    frame_pil: Image = frame.to_image()
+            # open video file for streaming
+            with self.av.open(str(video_file)) as container:
+                # get video stream
+                stream = container.streams.video[0]
 
-                    # use frame timestamp as image file name
-                    image_file_name = str(frame.time) + ".jpg"
+                # iterate frames
+                for frame_indx, frame in enumerate(container.decode(stream)):
+                    # check for k-th frame
+                    if not frame_indx % self.k:
+                        # get PIL image
+                        frame_pil: Image = frame.to_image()
 
-                    # save to output dir
-                    frame_pil.save(sample_sub_dir / image_file_name)
+                        # use frame timestamp as image file name
+                        image_file_name = str(frame.time) + ".jpg"
+
+                        # save to output dir
+                        frame_pil.save(sample_sub_dir / image_file_name)
 
 
 class Videolab:
@@ -67,28 +121,20 @@ class Videolab:
 
     def __init__(
         self,
-        video_dir: str,
+        video_dir: Optional[str] = None,
+        video_filepaths: Optional[List[str]] = None,
     ) -> None:
         """Create Path object from video directory string."""
-        # store video directory path
-        self.video_dir: Path = Path(video_dir)
-
-    def _find_videos(self) -> Generator[Path, None, None]:
-        """Iterate over video files in video directory."""
-        # iterate over video file extensions
-        for ext in VIDEO_FILE_EXTENSIONS:
-            # loop through video paths matching ext
-            yield from self.video_dir.glob(f"**/{ext}")
+        # store video dataset
+        self.video_dataset: VideoDataset = VideoDataset(video_dir, video_filepaths)
 
     def _sample_frames(self, samples_dir: Path, sample_interval: int) -> None:
         """Get sample frames."""
         # setup frame sampler
         frame_sampler = FrameSampler(sample_interval)
 
-        # iterate over video files in video data directory
-        for video_file in self._find_videos():
-            # sample frames from target video data directory
-            frame_sampler.sample(video_file, samples_dir)
+        # sample frames from target video data directory
+        frame_sampler.sample(self.video_dataset, samples_dir)
 
     def _parent_dir_frame_samples_dict(self) -> Dict[str, List[str]]:
         """Creates dictionary of parent directory and frame samples."""
--- a/src/cleanvision/videolab.py
+++ b/src/cleanvision/videolab.py
@@ -148,7 +148,7 @@ class Videolab:
 
     @staticmethod
     def _parent_dir_frame_samples_dict(
-        frame_issues: pd.Dataframe,
+        frame_issues: pd.DataFrame,
     ) -> Dict[str, List[str]]:
         """Creates dictionary of parent directory and frame samples."""
         # set dict
@@ -174,7 +174,7 @@ class Videolab:
         # get cluster dict
         return cluster_frame_samples
 
-    def _aggregate_issues(self, frame_issues: pd.Dataframe) -> pd.DataFrame:
+    def _aggregate_issues(self, frame_issues: pd.DataFrame) -> pd.DataFrame:
         """Aggregate Imagelab issues into a single frame for each video."""
         # convert booleans to floats
         pure_float_issues = frame_issues * 1
@@ -209,7 +209,7 @@ class Videolab:
         # return the aggregated dataframe
         return agg_df
 
-    def _aggregate_summary(self, aggregate_issues: pd.Dataframe) -> pd.DataFrame:
+    def _aggregate_summary(self, aggregate_issues: pd.DataFrame) -> pd.DataFrame:
         """Create issues summary for aggregate issues."""
         # setup issue summary storage
         summary_dict = {}
--- a/src/cleanvision/videolab.py
+++ b/src/cleanvision/videolab.py
@@ -52,7 +52,7 @@ class FrameSampler:
                     frame_pil.save(sample_sub_dir / image_file_name)
 
 
-class Videolab(Imagelab):
+class Videolab:
     """A single class to find all types of issues in video datasets."""
 
     def __init__(
@@ -86,7 +86,7 @@ class Videolab(Imagelab):
         cluster_frame_samples: Dict[str, List[str]] = {}
 
         # looper over index
-        for img_path in self.issues.index:
+        for img_path in self.imagelab.issues.index:
             # get frame sample parent
             sample_dir = Path(img_path).parents[0]
 
@@ -108,7 +108,7 @@ class Videolab(Imagelab):
     def _aggregate_issues(self) -> pd.DataFrame:
         """Aggregate Imagelab issues into a single frame for each video."""
         # convert booleans to floats
-        pure_float_issues = self.issues * 1
+        pure_float_issues = self.imagelab.issues * 1
 
         # store new aggregate_issues
         aggregate_issues = []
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -31,6 +31,7 @@ classifiers = [
 ]
 dependencies = [
     "Pillow>=9.3",
+    "av>=10.0.0",
```
jwmueller, line 34:
```suggestion
dependencies = [  # optional: av>=10.0.0
    "Pillow>=9.3",
```

src/cleanvision/videolab.py
```diff
@@ -0,0 +1,260 @@
+"""Videolab is an extension of Imagelab for finding issues in a video dataset."""
+from pathlib import Path
+from typing import Any, Dict, Generator, List, Optional
+
+import av
```
jwmueller, line 5:
```suggestion
```

pyproject.toml
jwmueller: suggest making this an optional dependency, so the package still works for image data without it installed

src/cleanvision/videolab.py
```diff
@@ -0,0 +1,260 @@
...
+
+        # return path to new sub dir
+        return sub_dir
+
+    def sample(self, video_file: Path, output_dir: Path) -> None:
+        """Loop through frames and store every k-th frame."""
+        # create frame samples sub directory
+        sample_sub_dir = self._create_frame_sample_sub_dir(video_file, output_dir)
+
+        # open video file for streaming
```
jwmueller, line 36:
```suggestion
        # open video file for streaming
        try:
            import av
        except ImportError as error:
            raise ImportError(
                "Cannot import package `av`. "
                "Please install it via `pip install av` and then try again."
             ) from error
        import av
```

jwmueller: Can lazy import the optional dependency `av` only where it's required

src/cleanvision/videolab.py
```diff
@@ -0,0 +1,260 @@
+"""Videolab is an extension of Imagelab for finding issues in a video dataset."""
```
jwmueller, line 1: amazing stuff!  

Could you comment:

A minimal example script of how to use this new Videolab and what the outputs look like?

A comment with comprehensive list of limitations of this code?
Eg: 
- What video file-type requirements are there. 
- What happens if some videos are really long some really short.
- any efficiency bottlenecks where the code feels slow
- edge cases where the results might not be good or code might crash.

lafemmephile: # Limitations
Exhaustive survey of the limitations of the current implementation for `Videolab` for *pull request #231*.

## General
The following is a brief *summary* of the general limitations of this pull request. For a more detailed (i.e. *low level*) analysis, see the **Specifics** section.

+ **NO** multiprocessing or multithreading used
+ **NO** default arguments used
+ **NO** parameter/error checking used
+ **NO** progress bars used (i.e. `tqdm`)
+ **NOT** all methods are implemented/overrided in `Videolab` class
+ **ONLY** these *video formats* are searched: https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L12

*Why?* From the discussion in #215, I decided it was best to minimize any *excess* decision making, since the `cleanvision` maintainers will have their own *standards* and *needs* for their library.

## Specifics
Here we will dig into the actual *code* and discuss the *select practical issues* that are currently present.

### Frame Sampling
Let us begin by looking at the *frame sampling* class aptly named `FrameSampler`:
https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L15-L52

#### Creating Frame Sample Directory: Collisions
One of the first real issues is in the *strategy* for naming the directories corresponding to the *frame samples* from a specific video file:
https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L22-L29

The current implementation uses the *user supplied* `output_dir` and the actual *file name* of the video. This may lead to situations where *collisions* occur: two video files have the same name, but different paths, and will map to the same *frame sample sub directory*. For example see the following diagram:
```console
/home/foo/cats/videos/001.mp4
               |
                \----> sample_frames/001.mp4/ <----\
                                                    |
                                        /home/foo/downloads/misc/memes/001.mp4
```

One way around this is by using some kind of *hash*, either on the *absolute file path* or the *video file contents*:
```console
/home/foo/cats/videos/001.mp4
               |
             HASH()
               |
                \----> sample_frames/9asd09kl23kl/001.mp4/

                sample_frames/jd82lkxja983/001.mp4/ <----\
                                                          |
                                                        HASH()
                                                          |
                                        /home/foo/downloads/misc/memes/001.mp4
```

Certainly there are multiple ways this could be done. The above shows one potential strategy.

### Videolab Class
Now let us dive into the actual `Videolab` class:
https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L55-L260

#### Constructor
One of the first real changes or *issues* with the `Videolab` class in comparison to `Imagelab` is the `__init__()` method:
+ `Videolab`: https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L58-L64
+ `Imagelab`: https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/imagelab.py#L118-L148

As you can see in the comparison above, `Videolab` does not use the `Imagelab` *constructor* but instead simply takes in a *directory path* as a `str` and stores it as a `pathlib.Path` object. This means that currently `Videolab` does not accept the other *dataset* parameters that `Imagelab` does. The reason why this was chosen will be reviewed in the **Finding Issues** section.

#### Frame Sampling: Single Process/Thread
One of the obvious performance issues will be in the actual *frame sampling*:
https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L73-L81

Whereas the original [VideoSampler](https://github.com/LemurPwned/video-sampler) written by @LemurPwned used multiprocessing, I chose to write a *minimal* version that does not. That way the *performance strategy* for the frame sampler can be dealt with separately in a manner that fits the maintainer's needs.

#### Aggregating Image Issues
One of the *unique* tasks that `Videolab` tackles is the *aggregation* of issues found
by `Imagelab`:
https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L108-L139

There may be some value in considering the *performance* of the *aggregation algorithm* which essentially store a `pandas.DataFrame` of the *mean* of all the *video frames* for a specific video in a `list`, which is finally *concatenated* with the `pandas.concat` function. This may or may not be the *optimal* way to aggregate issues.

Another point worth considering is what *sample frame* to use as the **"image"** that is representative of the video. Currently this line calculates which *sample frame* to use as a new index in the new aggregate issue:
https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L121-L122

The current strategy is to use *approximately* the *middle frame* to represent the video in the new *aggregate issues* `pandas.DataFrame`.

#### Finding Issues
The `cleanvision` maintainers may choose a different approach to how I have implemented `find_issues()`:
https://github.com/lafemmephile/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/videolab.py#L162-L182

Specifically they will notice that opted to execute the *frame sampling* step in the `find_issues` method, and then subsequently call the *constructor* for `Imagelab`, and then call the `Imagelab.find_issues` parent method.

lafemmephile: > A comment with comprehensive list of limitations of this code? Eg:
> 
>     * What video file-type requirements are there.
> 
>     * What happens if some videos are really long some really short.
> 
>     * any efficiency bottlenecks where the code feels slow
> 
>     * edge cases where the results might not be good or code might crash.

@jwmueller I will now look at developing some *tests* and *performance experiments* in this [repo](https://github.com/lafemmephile/videolab) to attempt to rigorously answer these questions.

lafemmephile: # Performance
Using the *Videolab demo* found [here](https://github.com/lafemmephile/videolab), the following `scalene` profiles were generated.

## Reduced Profile
Screenshot from 2023-10-23 14-41-31

## Full Profile
Screenshot 2023-10-23 at 14-46-53 Screenshot

lafemmephile: @jwmueller I've done some *preliminary* performance profiling using `scalene` (see above comment). Next I want to scale it up to a **MUCH LARGER** dataset (not just the measly 6 videos from the *demo) and do some *performance* profiling.

lafemmephile: # Performance: Videolab Demo 5K
Using the *Videolab demo* found [here](https://github.com/lafemmephile/videolab/blob/59be68966873057900936ecf3d90514cb9814a24/notebooks/videolab_demo.ipynb) (and available in *Google Colab* [here](https://colab.research.google.com/github/lafemmephile/videolab/blob/master/notebooks/videolab_demo.ipynb)), the following `scalene` profiles were generated. In the course of the *performance profiling* a total of **4999 videos** are downloaded (from the [WebVid-10M dataset](https://maxbain.com/webvid-dataset/)), sampled, and analyzed by `Imagelab`.

# Reduced Profile
```
scalene --reduced-profile --cli --html --outfile reduced_profile.html videolab_demo.py
```
Screenshot 2023-10-23 at 19-39-53 Screenshot

lafemmephile: @jwmueller From generating these `scalene` profiles, I can honestly say that the `FrameSampler` is horribly slow (something @LemurPwned and @smttsp were well aware of). For example, running the above `Videolab` on the *5k* sub-sample of the [WebVid-10M dataset](https://maxbain.com/webvid-dataset/)  took ~40 minutes to sample frames from 4999 videos, whereas `Imagelab` was able to process about *88k* images in ~10 minutes.

Hence, performance appears to be most affected by the *frame sampling strategy* than any other step in `Videolab`. Whether in this *pull request* or another separate issue, it will require a far more *optimized* and *intelligent* design than the simple `FrameSampler` that I have implemented (something closer to @LemurPwned's [VideoSampler](https://github.com/LemurPwned/video-sampler).

jwmueller: Great callout, ya replacing the FrameSampler will certainly need to be improved before we publicly announce this module to users.  But I don't think it is critical to this PR being merged and could be addressed in a future follow-up PR.

Thanks for the comprehensive benchmarking/profiling, very helpful!  Our team will get to the code review as soon as we're able to

LemurPwned: Great job, if you need me to optimise this, let me know, there are some avenues to explore here.

lafemmephile: @LemurPwned Thank you. It was a team effort by all of you. I just did what I could to *"lay the foundation"*. Yes definitely after this initial PR (like @jwmueller suggested) we can really optimize the *frame sampling*, and your and @smttsp's expertise in this matter are invaluable.

sanjanag:
Hi @lafemmephile ! This PR is a great start to implementing Videolab. I looked at the PR and here are my thoughts on it.

For the question of how to keep filenames for images, hashing, etc. I suggest we use a VideoDataset class, that maintains the mapping between numeric indices like [0,1,2,3] to video_paths. This would help us in supporting both `video_dir` and `video_filepaths` as arguments and reduce the overhead of hashmap, collisions etc. You can use this class a base class https://github.com/cleanlab/cleanvision/blob/43584bcb5554d52236d657d1fb9e7f1664eaf101/src/cleanvision/dataset/base_dataset.py

Using this index you can retrieve, video path, hence, video frames and the video itself. For saving frames, you could use the dir structure
```
frame_dir/

- 0/
    - frame_0.jpg
    - frame_1.jpg
- 1/
    - frame_0.jpg
    - frame_1.jpg
 ```

**Comment by sanjanag** [comment]
I think frame sampler is okay for now. We can take up the performance of frame sampler and also the strategy to sample frames in a separate PR, when we benchmark this on different datasets.

**Comment by sanjanag** [comment]
@lafemmephile Is there a specific reason you inherited Videolab from Imagelab? I think we could just have an `self.imagelab` object inside the `Videolab` class as we are using only `find_issues` method in frames. We can expose the same methods but there doesn't seem to be a need to inherit the Imagelab class.

**Comment by sanjanag** [comment]
@lafemmephile `Videolab`'s list of supported issues should be separate from the list of issues supported by `Imagelab`. Right now it also checks for near/exact and the current algo doesn't make sense for videos. So I'd suggest maintain a separate list of issues that Videolab supports and pass them in `imagelab.find_issues()`

**Comment by sanjanag** [comment]
For visualization of videos with issues, we could construct gifs from the extracted frames, I got the idea from [here](https://huggingface.co/docs/transformers/tasks/video_classification#visualize-the-preprocessed-video-for-better-debugging).

**Comment by sanjanag** [comment]
@lafemmephile for faster iteration over the code and checking if the issues make any sense, I suggest you create a test dataset for videos similar to the one [we have for images](https://github.com/cleanlab/cleanvision#quickstart). For starters just add one video of each issue type and see how it works.

**Comment by lafemmephile** [comment]
> @lafemmephile Is there a specific reason you inherited Videolab from Imagelab? I think we could just have an `self.imagelab` object inside the `Videolab` class as we are using only `find_issues` method in frames. We can expose the same methods but there doesn't seem to be a need to inherit the Imagelab class.

@sanjanag This is a good question. At the time, I thought that *subclassing* would help reduce the *re-implementation* of all the same methods found in `Imagelab` ... but yes I see that this may have been more trouble than it is worth. Moving forward I will start looking at refactoring to *composition* as opposed to *inheritance*.

> @lafemmephile `Videolab`'s list of supported issues should be separate from the list of issues supported by `Imagelab`. Right now it also checks for near/exact and the current algo doesn't make sense for videos. So I'd suggest maintain a separate list of issues that Videolab supports and pass them in `imagelab.find_issues()`

@sanjanag So basically no near/exact, just all the other issues should be supported?

**Comment by sanjanag** [comment]
> So basically no near/exact, just all the other issues should be supported?

That's correct @lafemmephile

**Comment by lafemmephile** [comment]
@sanjanag @jwmueller @LemurPwned Should I also implement all the public methods found in `Imagelab` or just stick with the existing methods I have in `Videolab`?

**Comment by sanjanag** [comment]
> @sanjanag @jwmueller @LemurPwned Should I also implement all the public methods found in `Imagelab` or just stick with the existing methods I have in `Videolab`?

It would be nice to have all the public methods in `Imagelab` as they apply to `Videolab` as well, but you could start with the ones you already have.

File: src/cleanvision/videolab.py (hunk 3 above)
Comment (lafemmephile) [reply src/cleanvision/videolab.py:36]: Before I commit these changes I must clarify: should the *import* be at the top of the source file with the other imports? Like this:
```python
"""Videolab is an extension of Imagelab for finding issues in a video dataset."""
from pathlib import Path
from typing import Any, Dict, Generator, List, Optional

 try:
    import av
except ImportError as error:
    raise ImportError(
        "Cannot import package `av`. "
        "Please install it via `pip install av` and then try again."
     ) from error
import av

import pandas as pd
from PIL.Image import Image

from cleanvision.imagelab import Imagelab
from cleanvision.utils.utils import get_is_issue_colname
```

What are the benefits of importing it directly in the `sample` method? This method will be called for each video. Will that not be a performance issue? Just unfamiliar with this strategy :sweat_smile: ...

jwmueller: You can move this line of code wherever is the first time a user would do something related to videolab. I.e. when they construct the object.

The reason I don't think this line of code can be at the top of the file is because the file will be imported with: import cleanvision

And thus this exception will be raised, even for users who didn't want to use Videolab.

So you can really put this code anywhere that makes it possible for users to run Imagelab without having av installed (which you can easily test yourself), while ensuring the exception gets raised as soon as such users try to run Videolab.

lafemmephile: Excellent, I see. I will do exactly that. Thank you for clarifying.

pyproject.toml
lafemmephile: So something like this:
```toml
dependencies = [
    "Pillow>=9.3",
    "numpy>=1.20.0",
    "pandas>=1.1.5",
    "tabulate>=0.8.3", # pandas optional dependency for .to_markdown()
    "imagehash>=4.2.0",
    "tqdm>=4.53.0",
    "matplotlib>=3.4",
    "fsspec>=2023.1.0", # latest compatible with Python 3.7
]

[project.optional-dependencies]
huggingface = ["datasets>=2.7.0"]
pytorch = ["torchvision>=0.12.0"]
azure = ["adlfs>=2022.2.0"] # latest compatible with Python 3.7
gcs = ["gcsfs>=2022.1.0"] # latest compatible with Python 3.7
s3 = ["s3fs>=2023.1.0"] # latest compatible with Python 3.7
av = ["av>=10.0.0"]

all = ["cleanvision[huggingface,pytorch,azure,gcs,s3,av]"]
```

Or just don't touch the `pyproject.toml` file at all?

sanjanag: This looks good, instead of `av`, you could use the `video` keyword.

lafemmephile:
> For visualization of videos with issues, we could construct gifs from the extracted frames, I got the idea from [here](https://huggingface.co/docs/transformers/tasks/video_classification#visualize-the-preprocessed-video-for-better-debugging).

@sanjanag @jwmueller I have implemented all the `Imagelab` *public methods* on `Videolab` except `visualize`. To be clear, you want this pull request to implement the *GIF* visualization strategy? I've already been looking into how to do it, just want to clarify the exact requirements of this pull request before I dive deep into implementing this :smile:.

src/cleanvision/videolab.py
jwmueller: @lafemmephile Do you want to make one of your notebooks here:
https://github.com/lafemmephile/videolab/tree/master/notebooks

the official quickstart tutorial for Videolab? 

Here's a GH issue detailing what we need for our quick-start tutorial:
https://github.com/cleanlab/cleanvision/issues/239

Should be a simple PR of one of your existing notebooks, which you can make separately from the Videolab source code PR in parallel, if you're interested!

lafemmephile: Definitely! I will work on this as well!

lafemmephile:
@jwmueller @sanjanag @LemurPwned I have done my best to implement all the *PUBLIC* methods ... there may be some attributes (e.g. `issues`) that should be implemented in order to match the *usage* and *API* of `Imagelab` ... The only remaining public method not implemented is `visualize`.

I find it difficult to be clear about how to implement the `visualize` method ... should we *"visualize"* videos? Should they be playable? Or should we visualize individual frames? From the research I did it does not some trivial to plot videos/GIFs in matplotlib **AND** have them be playable ... GIFs can be plotted but it does not seem obvious to me how to make them playable.

Until it becomes more clear what is the best strategy for visualizing the `Videolab` data I will hold off on attempting to implement the `visualize` method. If any of you have some experience/insight/ideas of how to approach this problem, please comment :+1:.

sanjanag:
> @jwmueller @sanjanag @LemurPwned I have done my best to implement all the _PUBLIC_ methods ... there may be some attributes (e.g. `issues`) that should be implemented in order to match the _usage_ and _API_ of `Imagelab` ... The only remaining public method not implemented is `visualize`.
> 
> I find it difficult to be clear about how to implement the `visualize` method ... should we _"visualize"_ videos? Should they be playable? Or should we visualize individual frames? From the research I did it does not some trivial to plot videos/GIFs in matplotlib **AND** have them be playable ... GIFs can be plotted but it does not seem obvious to me how to make them playable.
> 
> Until it becomes more clear what is the best strategy for visualizing the `Videolab` data I will hold off on attempting to implement the `visualize` method. If any of you have some experience/insight/ideas related to approaching this problem, please comment 👍.

Hi @lafemmephile ! That is some amazing work. Could you elaborate on what difficulty you are facing in making videos playable? And does playable means you click and play, or it plays as soon as it is plotted? It could be the case that it requires some frontend support from jupyter notebook.

We could start with frames, where we don't plot all the say `dark` frames from a video. But sample one frame per `dark` video.
I also think with either frames or GIFs, it is important to display some kind of time information where one can find it in the video. For example, say frame_10.png is sampled to visualize, then what instant in the video does it occur and what does does it occur in. These are some important information we want to show apart from the sampled frame.

sanjanag:
@lafemmephile We want to wrap up this PR soon, as I will be on PTO from second week of December and wouldn't be available during that time. Would it be okay if I take the reins of this PR from here? I think you have done some phenomenal contribution. Could you please tell if there are any pending in-progress contributions from your side?

lafemmephile:
Excellent.

PR #233, opened 2023-11-08
modify python source code example

developer0hye:
When I tried to run python example code, I got the below error without 'if __name__ == '__main__':'.

image


developer0hye:
Hi @sanjanag ! I signed CLA!

sanjanag (approved):

PR #234, opened 2023-11-08
Removed type ignore according to type check

sanjanag:


elisno (approved):
LGTM, since this change passes the black-formatter check in CI.

Issue #230, opened 2023-10-06
What are these errors and warning?


csv610:
File "/opt/homebrew/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/spawn.py", line 164, in get_preparation_data
    _check_not_importing_main()
  File "/opt/homebrew/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/spawn.py", line 140, in _check_not_importing_main
    raise RuntimeError('''
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.

        To fix this issue, refer to the "Safe importing of main module"
        section in https://docs.python.org/3/library/multiprocessing.html

sanjanag:
Hi @csv610 ! Could you give more context? What were you trying to run and stacktrace related to cleanvision code?

sanjanag:
Hi @csv610 ! This might be caused by the multiprocessing library.
Could you try executing the code inside of a main block like this
```python
if __name__ == "__main__":
    imagelab = Imagelab(data_path)
    imagelab.find_issues()
````

jwmueller:
FYI you can also try: `imagelab.find_issues(..., n_jobs = 1)` to disable multiprocessing entirely.

PR #235, opened 2023-11-13
Added faq page

sanjanag:
- Added FAQ page
- Fixed linking for fsspec_dataset class
- Reworded some of the text in Quickstart page.

Preview link: https://cleanvision--235.org.readthedocs.build/en/235/index.html


README.md
```diff
@@ -67,6 +66,7 @@ imagelab.report(issue_types=issue_types)
 - [Additional example notebooks](https://github.com/cleanlab/cleanvision-examples)
 - [Documentation](https://cleanvision.readthedocs.io/)
 - [Blog Post](https://cleanlab.ai/blog/cleanvision/)
+- [FAQ](FAQ.md)
```
jwmueller, line 70: the FAQ needs to be linked from the documentation side bar too right?

jwmueller: why have the FAQ be in the GH repo instead of in the documentation website?  Normally FAQ would be in the documentation website

FAQ.md
```diff
@@ -0,0 +1,66 @@
+# FAQ
+
+Answers to frequently asked questions about the [cleanvision](https://github.com/cleanlab/cleanvision) open-source package.
+
+The code snippets in this FAQ come from a fully executable notebook you can run via Colab or locally by downloading it [here](https://github.com/cleanlab/cleanlab/blob/master/docs/source/tutorials/faq.ipynb).
```
jwmueller, line 5: not sure what is going on here, looks this is just copy/pasted with no proofreading

FAQ.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+1. **What kind of machine learning tasks can I use cleanvision for?**
```
jwmueller, line 8:
```suggestion
1. **What kind of machine learning tasks can I use CleanVision for?**
```

jwmueller: try to be consistent with the case used for the text

FAQ.md
```diff
@@ -0,0 +1,66 @@
...
+imagelab = Imagelab(hf_dataset=dataset, image_key="image")
+```
+
+
+
+For more detailed usage instructions and examples, please refer to the [official documentation](https://cleanvision.readthedocs.io/).
+
+## Known issues
+
+1. **Encountering error from `multiprocessing` library while running cleanvision.**
```
jwmueller, line 34: is this a good name for the type of issue?

AFAIK this issue when it pops up makes little mention of multiprocess, and most users have no idea the root cause was multiprocessing. The way you phrased the FAQ is only useful to users who already know the root cause. Try to never write FAQs like that, instead write them based on the exact error message the users will see (or a generalized version that most users will encounter something similar to). 

Cf. how I renamed the original GH issue and I pinned it to make it more visible:

https://github.com/cleanlab/cleanvision/issues/230

FAQ.md
```diff
@@ -0,0 +1,66 @@
...
+The above issue is caused by multiprocessing module working differently for macOS and Windows platforms. A detailed discussion of the issue can be found [here](https://github.com/cleanlab/cleanlab/issues/159).
+A fix around this issue is to run cleanvision in the main namespace like this
+```python
+if __name__ == "__main__":
+
+    imagelab = Imagelab(data_path)
+    imagelab.find_issues()
+    imagelab.report()
+```
+OR use `n_jobs=1`
```
jwmueller, line 60:
```suggestion
OR use `n_jobs=1` to disable parallel processing:
```

jwmueller: always try to explain what a potential solution is changing in order to work around the problem. In this case, there is clear downside of this solution, so it is important to explain it is disabling parallel processing

FAQ.md
```diff
@@ -0,0 +1,66 @@
...
+
+Apart from plain image files, CleanVision also works with HuggingFace and Torchvision datasets. You can use the dataset objects as is with the `image_key` argument.
+
+```python
+imagelab = Imagelab(hf_dataset=dataset, image_key="image")
+```
+
+
+
+For more detailed usage instructions and examples, please refer to the [official documentation](https://cleanvision.readthedocs.io/).
```
jwmueller, line 30: why did you choose not to have the FAQ as part of the documentation?

jwmueller (changes requested):
had many questions

README.md
sanjanag: Including in the docs

FAQ.md
sanjanag: sorry about that, did that in a haste. Removing it

docs/source/faq.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. code-block:: python3
+
+    imagelab.find_issues(issue_types={"light": {}, "blurry": {}})
+
+
+3. **What dataset formats does CleanVision support?**
+
+
+Apart from plain image files, CleanVision also works with HuggingFace and Torchvision datasets. You can use the dataset objects as is with the ``image_key`` argument.
```
jwmueller, line 23:
```suggestion
Apart from plain image files stored locally or in the cloud, CleanVision also works with HuggingFace and Torchvision datasets. You can use the dataset objects as is with the ``image_key`` argument.
```

jwmueller (approved):
LGTM, just had one suggestion but go ahead and merge after considering that change

PR #249, opened 2024-01-29
HF datasets version fix

sanjanag:
CI fixes:
- Install different versions of datasets for different python versions. Can't use same version as datasets dropped support for python3.7 in version 2.14.0 and it breaks tests using datasets library.
- Use black formatting for jupyter notebooks as well
- Update versions of black and flake8 precommit hooks
- Ignore flake8 error E704 as it is conflicting with black formatting
- Add target versions for black in CI pipeline and only use `check` option and not modify the files.


elisno (approved):
LGTM.

PR #250, opened 2024-01-31
Added a min size of 1 to avoid zero width/length in resize

sanjanag:
Avoids zero division in Image.resize() call of blurry check. This is caused by extreme odd aspect ratios present in the dataset. Assigned a min value of 1 to width/height if they are less than 1.


jwmueller (approved):
lgtm

